PR #1, opened 2019-02-05
Add slsReceiver and slsDetector CMake targets to the export set (used by downstream project)

sdebionne:
At ESRF we use slsDetectorPackage as a git submodule in the Lima tree. The client (the lima plugin) creates and install a [CMake package](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html) but since it depends on the slsDetectorPacakages, target names have to be set when installing the targets (i.e. `slsDetectorShared`).

Also, with our use case, it would be great if the options are prefixed with `SLS_` (i.e. `SLS_USE_HDF5`). I can do that in a separate PR if you agree.

Side question, do you have any plan to support CMake packages? I can help you with this if you want.


thattil:
We have also prefixed the options with SLS_. The CMake packages can be supported for the refactor branch. We plan to release the 4.0.1 today.

sdebionne:
Thanks Dhanya! Let me know if you need help with the CMake Packages. I cannot find the v4.0.x tags on Github, are you tagging your versions elsewhere (or just forgot to `git push --tags`)?

thattil:
And thanks to you for the pull request! 

When we make releases, we usually just make a separate branch of it, tags are more for internal use. So you should be able to find the branch 4.0.1 :)

Yes, that would be of great help, if you could help us with the CMake Packages. If you have the time, could you please do a pull request in the refactor branch? Otherwise, we should be able to get to it later :)

PR #2, opened 2019-02-11
[WIP] Add CMake packages (and versions)

sdebionne:
This is work in progress, some targets are not exported properly yet.

The modifications include:

 - Use GNUInstallDirs module to get the install dir (but force libraries to `lib` for conda)
 - `project_version.cmake` will try to get the version from git refs (things like v1.20.5 are accepted but not `jungfrau3.0.0.9`)
- to make the package relocatable, one need to use `$<BUILD_INTERFACE:XXX>` and `$<BUILD_INTERFACE:XXX>` generators when specifying include directories (I whish this is simpler)

After `make install`, you should get `lib\cmake\slsdetectorpackage\slsdetectorpackage-targets.cmake` and the like, that are used by downstream projects with `find_dependency`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,23 @@
 cmake_minimum_required(VERSION 2.8)
 # cmake_minimum_required(VERSION 3.5)
 project(slsDetectorPackage)
+include(cmake/project_version.cmake)
+
+# Include additional modules that are used unconditionally
+include(GNUInstallDirs)
+
+# If conda build, always set lib dir to 'lib'
+if($ENV{CONDA_BUILD})
+  set(CMAKE_INSTALL_LIBDIR "lib")
+endif()
+
+# Set lower / upper case project names
+string(TOUPPER "${PROJECT_NAME}" PROJECT_NAME_UPPER)
+string(TOLOWER "${PROJECT_NAME}" PROJECT_NAME_LOWER)
+
+# Set targets export name (used by slsDetectorPackage and dependencies)
+set(TARGETS_EXPORT_NAME "${PROJECT_NAME_LOWER}-targets")
+#set(namespace "${PROJECT_NAME}::")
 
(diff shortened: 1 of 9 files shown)


sdebionne:
Are you using a special build of libzmq? I am asking, because the slsSupportLib provides `libzmq.a` and I wondered if we could use [FindZMQ](https://github.com/zeromq/lzmq/blob/master/cmake/FindZMQ.cmake) instead?

thattil:
Yes, the libzmq is developed by a colleague of ours, but it is a good idea to switch to findzmq instead. We will also look into the changes from this pull request and merge them soon.

sdebionne:
@thattil With the last commit slsSupportLib, slsDetectorShared and slsReceiverShared are exported. I still need to test with my LImA plugin though. Don't hesitate if you have any questions!

erikfrojdh:
@sdebionne Do you have any experience with finding zmq with Cmake? It looks like the cmake integration is not the best. At least for older versions. We would like it to work on Red Hat 7 without support from conda... 

Added FindZeroMQ.cmake from https://github.com/zeromq/azmq in the branch https://github.com/slsdetectorgroup/slsDetectorPackage/tree/findzmq

How tight do we need to specify the versions?

sdebionne:
@erikfrojdh find_package(ZeroMQ) should work without to much hassle if zmq is installed in a standard place, i.e. a place that is already in [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/v3.13/variable/CMAKE_PREFIX_PATH.html) -does your zmq comes from REHL package manager? `FindZeroMQ.cmake` could eventually be improved to export an interface so that target_link_library(zmq) is all that is needed to set the headers / libraries.

Any chance to update `cmake_minimum_required`  to say `3.1`? I think CMake (config-file) packages are supported from v3. I personally use CMake from Conda when the platform does not provide a version recent enough.

erikfrojdh:
Which version would be ideal for you? Discussed it quickly here and we have basically no objections... For me 3.5 seems sensible then i could run automatic test discovery for Catch2 among other things

PR #3, opened 2019-02-12
Add slsReceiver ctor overload

sdebionne:
I suggest to add a constructor overload of `slsReceiverUsers` (and `slsReceiver`) that takes the TCP/IP port in argument. IMHO, it is the responsibility of the application that uses the slsDetector libraries to parse the command line/config file and provide the argument.

Also I would recommend to use `std::unique_ptr<>` instead of raw pointers, see [Avoid calling new and delete explicitly](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r11-avoid-calling-new-and-delete-explicitly) for the rationale.

Tell me if I'm not targeting the right branch for these modifications!

### DIFF
--- a/slsReceiverSoftware/include/slsReceiver.h
+++ b/slsReceiverSoftware/include/slsReceiver.h
@@ -4,10 +4,9 @@
  * @short creates the UDP and TCP class objects
  ***********************************************/
 
+#include <memory>
 
-
-class slsReceiverTCPIPInterface;
-
+#include "slsReceiverTCPIPInterface.h"
 #include "sls_detector_defs.h"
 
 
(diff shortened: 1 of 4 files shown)


sdebionne:
`std::make_unique` is indeed C++14 (commented in the PR) and I did not know about `sls::make_unique`. Do you want me to patch it?

erikfrojdh:
If you want to sure go ahead! Then I merge the pull request.

sdebionne:
@erikfrojdh Here you go!

thattil:
Sounds good. The overloaded function is a good idea. Thanks Samuel!

PR #6, opened 2019-03-05
Fix typos

sdebionne:
A small PR mostly to check if it is the right way to work with you!


thattil:
All good :) Should I delete the fix-typos branch?

erikfrojdh:
Yes this is typically good to do. You should even see a link in the pull request to delete it.

thattil:
Alright. For next time then :)

PR #7, opened 2019-03-05
Fix sprintf warnings

sdebionne:
Reported by gcc 7.3.

Here, iostreams (`std::ostringstream`) would offer a better solution as the type of the (template) variables `arg` and `retval` are automatically deduced while `sprintf` uses a hardcoded format string.


erikfrojdh:
Do you have any experience with fmtlib?[https://github.com/fmtlib/fmt]( https://github.com/fmtlib/fmt ) this for me would be a nice way of formatting strings and output...

sdebionne:
@erikfrojdh Nope, I have never used {fmt}, this project looks interesting though, thanks for the heads-up. I used [Boost.Format](https://www.boost.org/doc/libs/1_69_0/libs/format/doc/format.html) as most of my former projects were based on Boost.

PR #18, opened 2019-05-03
Rxctb

thattil:
1. Connected some commands properly to the detector:
adcinvert,
extsamplingsrc,
extsampling

2. Removed special reciever compile option and all related ctb callbacks. Added commands rx_dbitlist (list of digital enable bits) and rx_dbitoffset (offset to cut off in dig data) in receiver.
Analog data list and offset is not yet included.


slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2666,6 +2677,143 @@ uint32_t slsDetector::getADCEnableMask() {
     return shm()->adcEnableMask;
 }
 
+void slsDetector::setADCInvert(uint32_t value) {
+    uint32_t arg = value;
+    FILE_LOG(logDEBUG1) << "Setting ADC Invert to 0x" << std::hex << arg << std::dec;
+    if (shm()->onlineFlag == ONLINE_FLAG) {
+        sendToDetector(F_SET_ADC_INVERT, &arg, sizeof(arg), nullptr, 0);
```
erikfrojdh: You can  use sendToDetector without specifying the sizes
e.g. sendToDetector(F_SET_ADC_INVERT, arg, nullptr)

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2666,6 +2677,143 @@ uint32_t slsDetector::getADCEnableMask() {
...
+    // copy size and vector to shm
+    shm()->rxDbitListSize = list.size();
+    std::copy(list.begin(), list.end(), shm()->rxDbitList);
+ 
+    int ret = FAIL;
+    if (shm()->rxOnlineFlag == ONLINE_FLAG) {
+        int fnum = F_SET_RECEIVER_DBIT_LIST;
+        int arg = list.size();
+        auto receiver =
+            sls::ClientSocket("Receiver", shm()->rxHostname, shm()->rxTCPPort);
```
erikfrojdh: I think it woudl be better to use the normal interface, meaning sendToReceiver? Is there any technical reason why you don't want to do this?

thattil: I used it as you suggested. I can't do it for the F_GET_RECEIVER_DBIT_LIST, as I need to get the size first before I get the array.

thattil:
The getReceiverdbitlist at the moment cannot use the standard one, as we need to get the size of the array first before we get the array.

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2741,25 +2741,11 @@ void slsDetector::setReceiverDbitList(std::vector<int> list) {
...
-            throw ReceiverError("Receiver " + std::to_string(detId) +
-                                " returned error: " + std::string(mess));
-        }
-    }
-    if (ret == FORCE_UPDATE) {
-        ret = updateCachedReceiverVariables();
+        int args[list.size() + 1];
+        args[0] = list.size();
+        std::copy(std::begin(list), std::end(list), args + 1);
+        sendToReceiver(F_SET_RECEIVER_DBIT_LIST, args, sizeof(args), nullptr, 0);
```
erikfrojdh: Here you also don't need to set the size:
sendToReceiver(F_SET_RECEIVER_DBIT_LIST, args, nullptr);
Should be enought

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2666,6 +2677,129 @@ uint32_t slsDetector::getADCEnableMask() {
     return shm()->adcEnableMask;
 }
 
+void slsDetector::setADCInvert(uint32_t value) {
+    uint32_t arg = value;
+    FILE_LOG(logDEBUG1) << "Setting ADC Invert to 0x" << std::hex << arg << std::dec;
+    if (shm()->onlineFlag == ONLINE_FLAG) {
+        sendToDetector(F_SET_ADC_INVERT, &arg, nullptr);
```
erikfrojdh: Remove the ref on arg, otherwise the sizeof will deduce the wrong size 
sendToDetector(F_SET_ADC_INVERT, arg, nullptr);
I will change the template later so it would be a compilation error

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2666,6 +2677,129 @@ uint32_t slsDetector::getADCEnableMask() {
...
+    }
+}
+
+uint32_t slsDetector::getADCInvert() {
+    uint32_t retval = -1;
+    FILE_LOG(logDEBUG1) << "Getting ADC Invert";
+    if (shm()->onlineFlag == ONLINE_FLAG) {
+        sendToDetector(F_GET_ADC_INVERT, nullptr, 0, &retval, sizeof(retval));
```
erikfrojdh: Also here no size and no ref on retval, the following line is enough
sendToDetector(F_GET_ADC_INVERT, nullptr, retval);

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2666,6 +2677,129 @@ uint32_t slsDetector::getADCEnableMask() {
...
+
+    // copy size and vector to shm
+    shm()->rxDbitListSize = list.size();
+    std::copy(list.begin(), list.end(), shm()->rxDbitList);
+ 
+    if (shm()->rxOnlineFlag == ONLINE_FLAG) {
+        int args[list.size() + 1];
+        args[0] = list.size();
+        std::copy(std::begin(list), std::end(list), args + 1);
+        sendToReceiver(F_SET_RECEIVER_DBIT_LIST, args, sizeof(args), nullptr, 0);
```
erikfrojdh: Same here no need for sizeof 
sendToReceiver(F_SET_RECEIVER_DBIT_LIST, args, nullptr);

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2666,6 +2677,129 @@ uint32_t slsDetector::getADCEnableMask() {
...
+    if (shm()->rxDbitListSize) {
+        retval.resize(shm()->rxDbitListSize);
+        std::copy(shm()->rxDbitList, shm()->rxDbitList + shm()->rxDbitListSize, std::begin(retval));
+    }
+
+    return retval;
+}
+
+int slsDetector::setReceiverDbitOffset(int value) {
+    int arg = value;
```
erikfrojdh: No need to copy value into arg. You can use value to send it directly
sendToReceiver(F_RECEIVER_DBIT_OFFSET, value, retval);

slsReceiverSoftware/src/DataProcessor.cpp
```diff
@@ -28,7 +28,7 @@ DataProcessor::DataProcessor(int ind, detectorType dtype, Fifo* f,
 		bool* dsEnable, bool* gpEnable, uint32_t* dr,
 		uint32_t* freq, uint32_t* timer,
 		bool* fp, bool* act, bool* depaden, bool* sm,
-		int* ct, int* cdo, int* cad) :
+		std::vector <int> * cdl, int* cdo, int* cad) :
```
erikfrojdh: Why do you use a pointer to a vector?

slsReceiverSoftware/src/DataProcessor.cpp
```diff
@@ -519,6 +501,45 @@ void DataProcessor::PadMissingPackets(char* buf) {
...
+	auto dest = result.data();
+	const int digOffset = FIFO_HEADER_NUMBYTES + sizeof(sls_receiver_header) + (*ctbAnalogDataBytes) + (*ctbDbitOffset);
+	auto source = (uint64_t*)(buf + digOffset);
+
+	// loop through digital bit enable vector
+    for (auto bi : (*ctbDbitList)) {
+		// loop through the frame digital data
+        for (auto ptr = source; ptr < (source + numSamples);) {
+			// extract destination in 64 bit batches
+            for (int i = 0; i != 64; ++i) {
```
erikfrojdh: Should this really be 64 and not the dynamic range?

slsReceiverSoftware/src/DataProcessor.cpp
```diff
@@ -519,6 +501,45 @@ void DataProcessor::PadMissingPackets(char* buf) {
 	}
 }
 
+/** ctb specific */
+void DataProcessor::RearrangeDbitData(char* buf) {
```
erikfrojdh: Overall in this function I think we could get away with less pointers and more RAII and pass by value. But we can leave this for refactoring

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -2666,6 +2677,129 @@ uint32_t slsDetector::getADCEnableMask() {
...
+ 
+    if (shm()->rxOnlineFlag == ONLINE_FLAG) {
+        int args[list.size() + 1];
+        args[0] = list.size();
+        std::copy(std::begin(list), std::end(list), args + 1);
+        sendToReceiver(F_SET_RECEIVER_DBIT_LIST, args, sizeof(args), nullptr, 0);
+    }
+}
+
+std::vector<int> slsDetector::getReceiverDbitList() {
```
erikfrojdh: Do you think we have more functions that will use this pattern? In that case we could add a general function that sends fnum, size, data ... to avoid having to do this every time. Or if this is the only place we can keep it.

slsDetectorSoftware/src/slsDetector.cpp
thattil: I had tried it earlier, but it doesnt work when it is an array

slsDetectorSoftware/src/slsDetector.cpp
thattil: Same comment as before.

slsReceiverSoftware/src/DataProcessor.cpp
thattil: So that if there are changes done to the vector in the tcp interface, it will reflect. The same with all the other pointers.

slsReceiverSoftware/src/DataProcessor.cpp
thattil: This is what we had decided earlier. The dynamic range for the digital bits in ctb is not known to receiver and will have to be used by the user when processing the file. So, these bits are just put together in 64 bit batches.

slsReceiverSoftware/src/DataProcessor.cpp
thattil: The pointer style was to do less work from the client interface (all the parameters are passed this way), but I had been thinking of changing that to make it more clean. This will have to be done at a different time and yes, it would be refactoring.

slsDetectorSoftware/src/slsDetector.cpp
thattil: Yes, but we can do this later as another pr, including ROI.

erikfrojdh: Sounds reasonable

slsReceiverSoftware/src/DataProcessor.cpp
erikfrojdh: Ah makes sense!

slsDetectorSoftware/src/slsDetector.cpp
erikfrojdh: Could you please try again. This should work. I have done it on many places
sendToReceiver(F_SET_RECEIVER_DBIT_LIST, args, nullptr);

But if you add a ref on args it won't work since the size will be wrong. 
sendToReceiver(F_SET_RECEIVER_DBIT_LIST, &args, nullptr);

erikfrojdh:
Could you just change the sendToReceiver to not use the size and then we merge this?

slsDetectorSoftware/src/slsDetector.cpp
thattil: I tried without the ref on args.
sendToReceiver(F_SET_RECEIVER_DBIT_LIST, args, nullptr);
And it doesnt compile.

PR #22, opened 2019-05-14
Using FixedCapacityContainer for rxDbitList

erikfrojdh:
Moving from vector to be able to send and receive with the standard functions. 
In a second stage we should probably store the values on the receiver side as FixedCapacityContainers as well.


slsDetectorSoftware/include/slsDetector.h
```diff
@@ -17,7 +17,7 @@ class multiSlsDetector;
 class ServerInterface;
 
 #define SLS_SHMVERSION 0x190503
```
thattil: Change date to reflect change in shared memory structure

erikfrojdh:
Ok, should be in now.

PR #26, opened 2019-05-29
Integration tests for the command line

erikfrojdh:
Adding the output stream as a parameter to multiSlsDetectorCommand allows for testing of the command line interface.

### DIFF
--- a/slsDetectorSoftware/include/multiSlsDetectorClient.h
+++ b/slsDetectorSoftware/include/multiSlsDetectorClient.h
@@ -48,22 +48,22 @@ class multiSlsDetectorClient {
         bool update = true;
         if (action_ == slsDetectorDefs::PUT_ACTION &&
             parser.n_arguments() == 0) {
-            std::cout << "Wrong usage - should be: " << parser.executable()
+            os << "Wrong usage - should be: " << parser.executable()
                       << "[id-][pos:]channel arg" << std::endl;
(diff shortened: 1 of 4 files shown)


slsDetectorSoftware/include/multiSlsDetectorClient.h
```diff
@@ -48,22 +48,22 @@ class multiSlsDetectorClient {
         bool update = true;
         if (action_ == slsDetectorDefs::PUT_ACTION &&
             parser.n_arguments() == 0) {
-            std::cout << "Wrong usage - should be: " << parser.executable()
+            os << "Wrong usage - should be: " << parser.executable()
```
mbrueckner-psi: Is it wise to name it 'os' ? what about 'outstream'?

erikfrojdh: I think that it's a fairly common convention to name the output stream variable os in C++, From the declaration of the member variable. std::ostream &os; it should be clear what we are referring to.

But I'll let Dhanya chip in as well and if it's a majority for a new name I'll change it.

thattil: The outstream is already declared with the name os in numerous instances in the package. In my opinion, this name doesn't warrant the effort to change it everywhere else in the package.

PR #27, opened 2019-05-31
set starting frame number of next acquisition for both jungfrau and e…

thattil:
…iger. firmware has not implemented a get, so workaround. tests included. frame number 0 not allowed due to Eiger. Eiger max frame is 48 bit, while jungfrau is 64 bit


thattil:
Once this is done, I would merge simulator branch to developer, which would also require some changes from this feature into it, ie. starting frame number for the virtual acquisition.

erikfrojdh:
Also looks good. Maybe you can make the get methods const but otherwise no comments. For the future we should think about how to structure the servers, beginning to be a lot of ifdefs....

slsDetectorSoftware/include/multiSlsDetector.h
```diff
@@ -622,6 +622,20 @@ class multiSlsDetector : public virtual slsDetectorDefs {
...
+     * @param detPos -1 for all detectors in  list or specific detector position
+     */
+    void setStartingFrameNumber(uint64_t value, int detPos = -1);
+
+     /**
+     * Get starting frame number for the next acquisition
+     * @param detPos -1 for all detectors in  list or specific detector position
+     * @returns starting frame number
+     */
+    uint64_t getStartingFrameNumber(int detPos = -1);
```
erikfrojdh: could be const

slsDetectorSoftware/include/multiSlsDetector.h
```diff
@@ -1785,9 +1799,9 @@ class multiSlsDetector : public virtual slsDetectorDefs {
     /**
      * Gets the current frame index of receiver
      * @param detPos -1 for all detectors in  list or specific detector position
-     * @returns current frame index of receiver
+     * @returns average of all current frame index of receiver
      */
-    int getReceiverCurrentFrameIndex(int detPos = -1);
+    uint64_t getReceiverCurrentFrameIndex(int detPos = -1);
```
erikfrojdh: also here...

slsDetectorSoftware/include/multiSlsDetector.h
thattil: Made the argument of setStartingFrameNumber const. Could not make getStartingFrameNumber const as it sends the arguments to the detector.

Regarding servers, I completely agree. It is very error prone as it is at the moment, which was also why I preferred C++. We can look into it later.

slsDetectorSoftware/include/multiSlsDetector.h
thattil: This also can't use a const qualifier.

PR #28, opened 2019-06-03
Simulator

thattil:


erikfrojdh:
Looks good to me. I would remove the binaries for the virtual servers since you now easily can build these using CMake. Then we keep the repo form growing in size...

PR #31, opened 2019-06-05
exceptions TCP

erikfrojdh:


slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -538,134 +464,73 @@ int slsReceiverTCPIPInterface::send_update(sls::ServerInterface2 &socket) {
...
-	memset(mess, 0, sizeof(mess));
+int slsReceiverTCPIPInterface::set_roi(Interface &socket) {
 	static_assert(sizeof(ROI) == 4*sizeof(int), "ROI not packed");
-	int narg = -1;
-	socket.read(&narg,sizeof(narg));
-
+	auto narg = socket.receive<int>();
 	std::vector <ROI> arg;
 	for (int iloop = 0; iloop < narg; ++iloop) {
 		ROI temp{};
```
thattil: Maybe use socket.receive(temp) here as well?

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -682,1580 +547,793 @@ int slsReceiverTCPIPInterface::set_roi(sls::ServerInterface2 &socket) {
...
+	std::string temp = sls::IpToInterfaceName(ip1);
+	if (temp == "none"){
+		throw RuntimeError("Failed to get ethernet interface or IP");
+	} else {
+		char eth[MAX_STR_LENGTH]{};
+		strcpy(eth, temp.c_str());
+		// if there is a dot in eth name
+		if (strchr(eth, '.') != nullptr) {
+			strcpy(eth, "");
+			ret = FAIL;
```
thattil: This was there previously, but I now think "ret = FAIL" should be removed here and mac address should always be checked even if eth is blank. Then you can remove the following if (ret!=FAIL) for mac address and sendResult like the other functions. The same for the second part for jungfrau.

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -682,1580 +547,793 @@ int slsReceiverTCPIPInterface::set_roi(sls::ServerInterface2 &socket) {
...
-	// get args, return if socket crashed, ret is fail if receiver is not null
-	socket.receiveArg(ret, mess, &index, sizeof(index));
-	if(receiver == nullptr){
-		NullObjectError(ret, mess);
-		return FAIL;
-	}
-
-	// base object not null
-	if (ret == OK) {
+	socket.receiveArg(index);
```
thattil: Could you use socket.receive? or is it not possible since it is an array?

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -682,1580 +547,793 @@ int slsReceiverTCPIPInterface::set_roi(sls::ServerInterface2 &socket) {
...
+			throw RuntimeError(mess);
+			break;
+		}
+		retval=impl()->getNumberofAnalogSamples();
+		break;
+	case DIGITAL_SAMPLES:
+		if (myDetectorType != CHIPTESTBOARD && myDetectorType != MOENCH) {
+			sprintf(mess,"This timer mode (%lld) does not exist for this receiver type\n", (long long int)index[0]);
+			throw RuntimeError(mess);
+			break;
```
thattil: is break needed?

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -682,1580 +547,793 @@ int slsReceiverTCPIPInterface::set_roi(sls::ServerInterface2 &socket) {
...
+		retval=impl()->getSubExpTime();
+		break;
+	case SUBFRAME_DEADTIME:
+		retval=(impl()->getSubPeriod() - impl()->getSubExpTime());
+		break;
+	case ANALOG_SAMPLES:
+		if (myDetectorType != CHIPTESTBOARD && myDetectorType != MOENCH) {
+			sprintf(mess,"This timer mode (%lld) does not exist for this receiver type\n", (long long int)index[0]);
+			throw RuntimeError(mess);
+			break;
```
thattil: is break needed for compiler warnings?

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -682,1580 +547,793 @@ int slsReceiverTCPIPInterface::set_roi(sls::ServerInterface2 &socket) {
...
+	FILE_LOG(logDEBUG1) << "Reset frames caught";
+	impl()->resetAcquisitionCount();
+	return socket.sendData(OK);
+}
+
+int slsReceiverTCPIPInterface::enable_file_write(Interface &socket){
+	auto enable = socket.receive<int>();
+	// set
+	if (enable >= 0) {
+		// verify if receiver is unlocked and idle
```
thattil: minor. you can remove this comment as it does not check unlocked here

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -682,1580 +547,793 @@ int slsReceiverTCPIPInterface::set_roi(sls::ServerInterface2 &socket) {
...
-		}
-		// get
-        retval = receiver->getUDPSocketBufferSize();
-        if (index != 0)
-        	validate(index, retval, std::string("set udp socket buffer size (No CAP_NET_ADMIN privileges?)"), DEC);
-        FILE_LOG(logDEBUG1) << "UDP Socket Buffer Size:" << retval;
-	}
-	return socket.sendResult(true, ret, &retval, sizeof(retval), mess);
+int slsReceiverTCPIPInterface::get_real_udp_socket_buffer_size(Interface &socket){
+	return socket.sendResult(impl()->getActualUDPSocketBufferSize());
```
thattil: minor. logdebug print?

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -682,1580 +547,793 @@ int slsReceiverTCPIPInterface::set_roi(sls::ServerInterface2 &socket) {
...
-	}
-	return socket.sendResult(true, ret, &retval, sizeof(retval), mess);
+	FILE_LOG(logDEBUG1) << "ADC enable mask retval: " << retval;
+	return socket.sendResult(retval);
 }
 
-int slsReceiverTCPIPInterface::set_dbit_list(sls::ServerInterface2 &socket) {
-    ret = OK;
-    memset(mess, 0, sizeof(mess));
+int sls
```
thattil: throw since it is a fail?

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -209,234 +218,151 @@ int slsReceiverTCPIPInterface::function_table(){
...
-	socket.sendResult(false, ret, nullptr, 0, nullptr);
+	socket.sendData(OK);
 	return GOODBYE;
 }
 
-int slsReceiverTCPIPInterface::lock_receiver(sls::ServerInterface2 &socket) {
-    ret = OK;
-    memset(mess, 0, sizeof(mess));
-    int lock = 0;
+void slsReceiverTCPIPInterface::ThrowNullObjectError(Interface &socket){
```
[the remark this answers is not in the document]
erikfrojdh: Thanks, removing it

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
```diff
@@ -538,134 +464,73 @@ int slsReceiverTCPIPInterface::send_update(sls::ServerInterface2 &socket) {
...
-		return FAIL;
-
+int slsReceiverTCPIPInterface::set_detector_type(Interface &socket){
+	auto arg = socket.receive<detectorType>();
 	// set
 	if (arg >= 0) {
 		// if object exists, verify unlocked and idle, else only verify lock (connecting first time)
-		if (receiver == nullptr){
-			VerifyLock(ret, mess);
+		if (receiver != nullptr){
```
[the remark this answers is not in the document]
erikfrojdh: For now I need it since verify idle would throw when you don't have the receiver. We could rewrite it to pass if the receiver is not there...

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
erikfrojdh: Well spotted!

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
erikfrojdh: I'll trust you on this one

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
erikfrojdh: I would need to rewrite it to be an std::array<int64_t,2> instead of int64_t index[2]. 
The one that we have now decays to a pointer

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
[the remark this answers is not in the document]
thattil: yes, that is true.

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
erikfrojdh: should be fine, I'll remove them

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
erikfrojdh: done!

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
erikfrojdh: But I like one liners =) 
Anyway fixed it. Later we can think about how many debug messages we need....

slsReceiverSoftware/src/slsReceiverTCPIPInterface.cpp
erikfrojdh: I removed the check completely since we already check in the slsDetector

thattil: Yes, you are right.

PR #38, opened 2019-06-28
New filename structure

erikfrojdh:
The new structure is:

[name]_d[detector index]_f[file index]_[measurement index]

example run_d0_f0_0.raw

On overflow f is increased

example run_d0_f0_0.raw
example run_d0_f1_0.raw
example run_d0_f2_0.raw

On the next measurement the measurement index is increased

example run_d0_f0_0.raw
...
example run_d0_f0_1.raw

### DIFF
--- a/slsReceiverSoftware/include/BinaryFileStatic.h
+++ b/slsReceiverSoftware/include/BinaryFileStatic.h
@@ -41,40 +41,33 @@ class BinaryFileStatic {
 	 * @param unitindex unit index
 	 * @returns complete file name created
 	 */
-	static std::string CreateFileName(char* fpath, char* fnameprefix, uint64_t findex, bool frindexenable,
-			uint64_t fnum = 0, int dindex = -1, int numunits = 1, int unitindex = 0, bool fixedw_findex = false)
-	{
-		std::ostringstream osfn;
-		osfn << fpath << "/" << fnameprefix;
-		if (dindex >= 0) osfn << "_d" << (dindex * numunits + unitindex);
-		if (frindexenable) osfn << "_f" << std::setfill('0') << std::setw(12) << fnum;
-			osfn << "_";
-		if (fixedw_findex)
-			osfn << std::setfill('0') << std::setw(12) << findex;
-		else
-			osfn << findex;
-		osfn << ".raw";
-		return osfn.str();
+	static std::string CreateFileName(char *fpath, char *fprefix,
+										uint64_t findex, uint64_t fnum,
+										int dindex, int numunits = 1,
+										int unitindex = 0) {
+		std::ostringstream os;
+		os << fpath << "/" << fprefix << "_d"
+			<< (dindex * numunits + unitindex) << "_f" << fnum << '_'
+			<< findex << ".raw";
+		return os.str();
 	}
 
-	/**
+        /**
 	 * Create file names for master file
 	 * @param fpath file path
 	 * @param fnameprefix file name prefix (includes scan and position variables)
 	 * @param findex file index
 	 * @returns master file name
 	 */
-	static std::string CreateMasterFileName(char* fpath, char* fnameprefix, uint64_t findex)
-	{
-		std::ostringstream osfn;
-		osfn << fpath << "/" << fnameprefix;
-		osfn << "_master";
-		osfn << "_" << findex;
-		osfn << ".raw";
-		return osfn.str();
+	static std::string CreateMasterFileName(char *fpath, char *fnameprefix,
+											uint64_t findex) {
+		std::ostringstream os;
+		os << fpath << "/" << fnameprefix << "_master"
+			<< "_" << findex << ".raw";
+		return os.str();
 	}
 
-	/**
+        /**
 	 * Close File
 	 * @param fd file pointer
 	 */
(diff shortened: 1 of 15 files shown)


slsReceiverSoftware/include/BinaryFileStatic.h
```diff
@@ -40,36 +41,33 @@ class BinaryFileStatic {
 	 * @param unitindex unit index
```
thattil: I would correct the comment to say it is the sub frame index and not frame number index. I would also remove frindexenable from the comment. I would use "subfIndex" rather than fnum as a variable name.

slsReceiverSoftware/include/HDF5FileStatic.h
```diff
@@ -47,19 +47,18 @@ class HDF5FileStatic: public virtual slsDetectorDefs {
 	 * @param unitindex unit index
```
thattil: I would correct the comment to say it is the sub frame index and not frame number index. I would also remove frindexenable from the comment. I would use "subfIndex" rather than fnum as a variable name.

slsReceiverSoftware/include/HDF5FileStatic.h
```diff
@@ -675,7 +672,7 @@ class HDF5FileStatic: public virtual slsDetectorDefs {
 
 				//source file name
 				std::string srcFileName = HDF5FileStatic::CreateFileName(fpath, fnameprefix, findex,
-						frindexenable, framesSaved, dindex, numunits, i);
+						framesSaved, dindex, numunits, i);
```
thattil: framesSaved shouldn't be used to create the file name anymore. It should be the hyperslab loop variable "j" most likely. It looks like you test hdf5, but I don't understand how it found the source files to create the virtual one. Did it create the virtual files then?

slsReceiverSoftware/src/BinaryFile.cpp
```diff
@@ -72,7 +72,7 @@ int BinaryFile::WriteToFile(char* buffer, int buffersize, uint64_t fnum, uint32_
 	// check if maxframesperfile = 0 for infinite
 	if ((*maxFramesPerFile) && (numFramesInFile >= (*maxFramesPerFile))) {
 		CloseCurrentFile();
-		CreateFile(fnum);
+		CreateFile(++subFileIndex);
```
thattil: You don't need to pass the subfleindex as it is accessible by CreateFile. I believe you just need to increment it in CreateFile, but then you will have to reset it to -1 (or increment it here and call CreateFile). Please also edit the comment in the header file for CreateFile stating what fnum is as it is misleading.

slsReceiverSoftware/src/DataProcessor.cpp
```diff
@@ -252,11 +252,12 @@ int DataProcessor::CreateNewFile(bool en, uint64_t nf, uint64_t at, uint64_t st,
 	if (file == nullptr)
 		return FAIL;
 	file->CloseAllFiles();
+	file->resetSubFileIndex();
 	if (file->CreateMasterFile(*masterFileWriteEnable, en,	generalData->imageSize,
 			generalData->nPixelsX, generalData->nPixelsY,
 			at, st, sp, ap) == FAIL)
 		return FAIL;
-	if (file->CreateFile(currentFrameIndex) == FAIL)
+	if (file->CreateFile(0) == FAIL)
```
thattil: You do not need to pass the 0 as it should already know subframeindex value.

slsReceiverSoftware/src/HDF5File.cpp
```diff
@@ -199,7 +199,7 @@ int HDF5File::WriteToFile(char* buffer, int buffersize, uint64_t fnum, uint32_t
 	// check if maxframesperfile = 0 for infinite
 	if ((*maxFramesPerFile) && (numFramesInFile >= (*maxFramesPerFile))) {
 		CloseCurrentFile();
-		CreateFile(fnum);
+		CreateFile(++subFileIndex);
```
thattil: Same as in BinaryFile.cpp
You don't need to pass the subfleindex as it is accessible by CreateFile. I believe you just need to increment it in CreateFile, but then you will have to reset it to -1 (or increment it here and call CreateFile). Please also edit the comment in the header file for CreateFile stating what fnum is as it is misleading.

slsReceiverSoftware/include/BinaryFileStatic.h
erikfrojdh: ok, done

slsReceiverSoftware/include/HDF5FileStatic.h
erikfrojdh: ok, done

slsReceiverSoftware/src/BinaryFile.cpp
erikfrojdh: Ok, since the variable is unsigned I went with the increment option, even though overflow would have worked.

slsReceiverSoftware/src/DataProcessor.cpp
erikfrojdh: ok, done

slsReceiverSoftware/src/HDF5File.cpp
erikfrojdh: ok, done

slsReceiverSoftware/include/HDF5FileStatic.h
erikfrojdh: Good catch! changed it to j

PR #39, opened 2019-07-22
Quad

thattil:
- Quad for developer branch (still has a hardcoded value for old Eiger Enum in multislsdetector until eiger fimware updates.
- detector shape and gap pixels enable added to json header instead of picking it up from shm.
- bug fix (gui) for hidden trimbits menu when in experts mode


erikfrojdh:
Looks good to me, if you tested it please go ahead and merge. 
However, I might have missed something since I don't know this part of the code.

PR #41, opened 2019-07-23
Removed multiSlsDetector dependency from slsDetector

erikfrojdh:
- Removed multiSlsDetector dependency from slsDetector by moving writing of config file to multi.
- Removed static version of freeSharedMemory from slsDetector since all should go through multi

### DIFF
--- a/slsDetectorSoftware/src/slsDetector.cpp
+++ b/slsDetectorSoftware/src/slsDetector.cpp
@@ -37,7 +37,7 @@ slsDetector::slsDetector(detectorType type, int multi_id, int det_id,
         FILE_LOG(logWARNING) << "This shared memory should have been "
                                 "deleted before! "
                              << shm.GetName() << ". Freeing it again";
-        freeSharedMemory(multi_id, det_id);
+       shm.RemoveSharedMemory();
     }
 
     initSharedMemory(type, multi_id, verify);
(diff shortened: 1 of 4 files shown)


slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -948,124 +940,80 @@ int slsDetector::updateDetector() {
...
-        names.emplace_back("rx_hostname");
-
-        names.emplace_back("extsig:0");
-        names.emplace_back("vhighvoltage");
+        base.emplace_back("detectormac");
+        base.emplace_back("detectorip");
+        base.emplace_back("rx_udpport");
+        base.emplace_back("rx_udpip");
+        base.emplace_back("rx_udpmac");
+        base.emplace_back("extsig:0");
```
thattil: For extsig, we removed the signal index. So, just "extsig" should do.

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -948,124 +940,80 @@ int slsDetector::updateDetector() {
...
-        names.emplace_back("rx_hostname");
-
-        names.emplace_back("powerchip");
-        names.emplace_back("vhighvoltage");
+        base.emplace_back("detectormac");
+        base.emplace_back("detectorip");
+        base.emplace_back("rx_udpport");
+        base.emplace_back("rx_udpip");
+        base.emplace_back("rx_udpmac");
+        base.emplace_back("powerchip");
```
thattil: This is a bug I introduced sometime back. Could you please remove the powerchip command for Moench?

slsDetectorSoftware/src/slsDetector.cpp
erikfrojdh: done

slsDetectorSoftware/src/slsDetector.cpp
erikfrojdh: done

PR #47, opened 2019-08-06
Partialreadout

thattil:
partial readout for eiger
tests (removed setonline, getreceiveronline)
gui (up and down in Terminal tab picks up last command and clears the current command)


slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -607,6 +607,23 @@ void slsDetector::setQuad(const bool enable) {
...
+void slsDetector::setReadNLines(const int value) {
+    FILE_LOG(logDEBUG1) << "Setting read n lines to " << value;
+    sendToDetector(F_SET_READ_N_LINES, value, nullptr);
+    FILE_LOG(logDEBUG1) << "Setting read n lines to " << value << " in Receiver";
+    if (shm()->useReceiverFlag) {
+        sendToReceiver(F_SET_RECEIVER_READ_N_LINES, value, nullptr);
+}
+}
+
+int slsDetector::getReadNLines() {
```
erikfrojdh: Philosophical question: Do we need both log lines? In this function we have two lines of logging and three lines of actual code... maybe it's possible to move the logging into sendToDetector?

slsDetectorServers/slsDetectorServer/slsDetectorFunctionList.h
```diff
@@ -252,8 +252,10 @@ int 		setDetectorPosition(int pos[]);
 #ifdef EIGERD
 int			setQuad(int value);
 int			getQuad();
-int		setInterruptSubframe(int value);
+int			setInterruptSubframe(int value);
 int			getInterruptSubframe();
+int 		setReadNLines(int value);
```
erikfrojdh: Is it only on github the formatting is off?

slsReceiverSoftware/include/receiver_defs.h
```diff
@@ -36,7 +36,7 @@
 
 //parameters to calculate fifo depth
 #define SAMPLE_TIME_IN_NS				100000000//100ms
-
+#define MAX_EIGER_ROWS_PER_READOUT		(256)
```
erikfrojdh: Are the () needed? Also maybe we can run format on this file, seems like the alignment is off...

slsDetectorServers/slsDetectorServer/slsDetectorFunctionList.h
thattil: It doesnt look so on my editor, but I haven't run clang on any of the servers anyway. Will do the formatting at a later step.

slsReceiverSoftware/include/receiver_defs.h
thattil: It doesn't need it and again had looked fine with the editor, but will do the formatting for all the files at a later step.

thattil: done with branckets

PR #50, opened 2019-08-21
Less data race in slsReceiver

erikfrojdh:
Changed some variables to std::atomic to avoid data races. 
Further work is needed but this is a first step

### DIFF
--- a/slsSupportLib/include/genericSocket.h
+++ b/slsSupportLib/include/genericSocket.h
@@ -43,7 +43,7 @@ class sockaddr_in;
 #include <errno.h>
 #include <stdio.h>
 #include "logger.h"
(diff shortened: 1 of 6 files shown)


slsSupportLib/include/genericSocket.h
```diff
@@ -43,7 +43,7 @@ class sockaddr_in;
 #include <errno.h>
 #include <stdio.h>
 #include "logger.h"
-
+#include <atomic>
```
thattil: minor. but you probably dont need this here.

erikfrojdh: Well spotted, removed it now.

PR #54, opened 2019-08-26
Guidetector

thattil:
Gui updated to detector.h and chrono


slsDetectorGui/src/qDetectorMain.cpp
```diff
@@ -116,10 +116,10 @@ qDetectorMain::qDetectorMain(int argc, char **argv, QApplication *app,
 }
 
 qDetectorMain::~qDetectorMain() {
-    if (myDet)
-        delete myDet;
-    if (myPlot)
-        delete myPlot;
+    if (det)
```
erikfrojdh: You don't have to check for null pointer before deleting the object. If it is a null pointer the
delete does nothing. Only when reading or writing to a pointer you need to check.

slsDetectorGui/src/qDetectorMain.cpp
```diff
@@ -144,22 +144,22 @@ void qDetectorMain::SetUpWidgetWindow() {
...
 
     // creating all the other tab widgets
-    tabMeasurement = new qTabMeasurement(this, myDet, myPlot);
-    tabDataOutput = new qTabDataOutput(this, myDet);
-    tabPlot = new qTabPlot(this, myDet, myPlot);
-    tabSettings = new qTabSettings(this, myDet);
-    tabAdvanced = new qTabAdvanced(this, myDet);
-    tabDebugging = new qTabDebugging(this, myDet);
-    tabDeveloper = new qTabDeveloper(this, myDet);
+    tabMeasurement = new qTabMeasurement(this, det, plot);
```
erikfrojdh: Using unique pointers instead of normal pointers then you would not need to delete the object manually in the destructor.

std::uniquie_ptr<qTabMeasurement> = sls::make_unique<qTabMeasurement>(this, det, plot) 
should work.

slsDetectorGui/src/qDetectorMain.cpp
```diff
@@ -242,30 +241,27 @@ void qDetectorMain::SetUpDetector(const std::string fName, int multiID) {
...
     case slsDetectorDefs::MOENCH:
-        actionLoadTrimbits->setEnabled(false);
-        actionSaveTrimbits->setEnabled(false);
-        break;
+         break;
     default:
         std::string errorMess =
-            myDet->getHostname() + std::string(" has ") +
-            myDet->getDetectorTypeAsString() + std::string(" detector type (") +
+            sls::ToString(det->getHostname(), ',') + std::string(" has ") +
```
erikfrojdh: This should also work without the std::string() around the string litterals if at least one object is a string. 

 sls::ToString(det->getHostname(), ',') +  has " + ...

should work and is a bit cleaner.

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -768,7 +768,7 @@ void Detector::setRxAddGapPixels(bool enable) {
 Result<bool> Detector::getParallelMode(Positions pos) const {
     auto res = pimpl->Parallel(&slsDetector::setReadOutFlags, pos,
                                defs::GET_READOUT_FLAGS);
-    Result<bool> booleanRes;
+    Result<bool> booleanRes(res.size());
     for (unsigned int i = 0; i < res.size(); ++i) {
```
erikfrojdh: Maybe use size_t here?

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -768,7 +768,7 @@ void Detector::setRxAddGapPixels(bool enable) {
...
         booleanRes[i] = (res[i] & defs::PARALLEL) ? true : false;
```
erikfrojdh: This code should be equivalent. Since both versions rely on implicit conversion. 
booleanRes[i] = res[i] & defs::PARALLEL;

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -250,14 +250,14 @@ void Detector::setTimingMode(defs::timingMode value, Positions pos) {
 void Detector::acquire() { pimpl->acquire(); }
 
 void Detector::startAcquisition() {
-    if (getUseReceiverFlag({}).squash())
+    if (getUseReceiverFlag({}).squash(true))
```
erikfrojdh: Do you need the {}  in the function call?

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -250,14 +250,14 @@ void Detector::setTimingMode(defs::timingMode value, Positions pos) {
...
         pimpl->Parallel(&slsDetector::startReceiver, {});
     pimpl->Parallel(&slsDetector::startAcquisition, {});
 }
 
 void Detector::stopAcquisition() {
     pimpl->Parallel(&slsDetector::stopAcquisition, {});
-    if (getUseReceiverFlag({}).squash()) // TODO: problem for acquire()
+    if (getUseReceiverFlag({}).squash(true))
```
erikfrojdh: Same here...

slsDetectorGui/src/qDetectorMain.cpp
thattil: done

slsDetectorSoftware/src/Detector.cpp

slsDetectorSoftware/src/Detector.cpp

slsDetectorSoftware/src/Detector.cpp
thattil: No. Good catch. Removed them.

slsDetectorSoftware/src/Detector.cpp
thattil: I removed it for the  getUseReceiverFlag. But I need them for the Parallel call after stopAcquisition. To remove it from there, I need to edit multiSlsDetector::Parallel to allow default {} for positions. Shall I do that? Or was it omitted for some reason?

slsDetectorGui/src/qDetectorMain.cpp
thattil: done

slsDetectorGui/src/qDetectorMain.cpp
thattil: done

PR #59, opened 2019-08-28
adc invert for highz

thattil:


erikfrojdh:
Looks good, my only comment is if this is something that should be done in firmware?

PR #60, opened 2019-08-28
rx_lock and lock back on, already in detector.h

thattil:


erikfrojdh:
Sorry I was a bit distracted, lock and rx_lock has already been merged to the new API. You find it in the list if you do: sls_detector_get list migrated or sls_detector_get list | grep lock.

PR #61, opened 2019-08-30
Readoutflags

thattil:


thattil:
Readout flags and speedLevel (maybe we can change this enum name)

PR #71, opened 2020-01-14
Counters

thattil:
mythen3
1. set and get counter mask (only number of counters implemented in firmware for the time being)
2. check type of module attached before powering on chip


slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -3888,6 +3865,32 @@ void slsDetector::setPipeline(int clkIndex, int value) {
...
 }
 
+void slsDetector::setCounters(uint32_t countermask) {
+    FILE_LOG(logDEBUG1) << "Setting Counters to " << countermask;
+    sendToDetector(F_SET_COUNTERS, countermask, nullptr);
+    sendCounterstoReceiver(countermask);
+}
+
+void slsDetector::sendCounterstoReceiver(uint32_t countermask) {
+    if (shm()->useReceiverFlag) {
```
erikfrojdh: Speedwise it's not needed but just for fun, many CPUs come with a special instruction to count bits in an integer. This is accessible through int __builtin_popcount (unsigned int x) on both clang and gcc. 

Just for fun you can compare the length of the assembly code. 
[https://godbolt.org/z/QuY4Xu](https://godbolt.org/z/QuY4Xu)

Here there is no speed advantage but replacing the loop with the popcount function means less code to worry about.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -305,6 +305,8 @@ const char* getFunctionName(enum detFuncs func) {
 	case F_SET_BURST_MODE:					return "F_SET_BURST_MODE";
 	case F_SET_ADC_ENABLE_MASK_10G:         return "F_SET_ADC_ENABLE_MASK_10G";
 	case F_GET_ADC_ENABLE_MASK_10G:         return "F_GET_ADC_ENABLE_MASK_10G";
+	case F_SET_COUNTERS:         			return "F_SET_COUNTERS";
```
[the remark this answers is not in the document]
erikfrojdh: but again maybe better to keep in since it's then closer to the function

slsDetectorSoftware/include/Detector.h
```diff
@@ -910,7 +910,18 @@ class Detector {
...
-    Result<bool> getBurstMode(Positions pos = {});    
+    Result<bool> getBurstMode(Positions pos = {});   
+
+    /**************************************************
+     *                                                *
+     *    Mythen3 Specific                            *
+     *                                                *
+     * ************************************************/
+    /** [Mythen3] */
+    Result<uint32_t> getCounters(Positions pos = {}) const;
```
erikfrojdh: getCounterMask?

slsDetectorServers/mythen3DetectorServer/slsDetectorFunctionList.c
```diff
@@ -529,6 +529,82 @@ int64_t getPeriod() {
...
 }
 
+void setCounters(uint32_t arg) {
+	if (arg == 0 || arg > COUNTER_MSK) {
+		return;
+	}
+	countermask = arg;
+	// convert mask into number of counters (until firmware converts to mask)
+	int ncounters = 0;
+	{
```
erikfrojdh: maybe also popcount?

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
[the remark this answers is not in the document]
thattil: Makes sense. Changed API to setCounterMask and getCounterMask. Leaving command line to counters as it is still a list. If we come up with a better command line name, we can change it.

slsDetectorSoftware/src/slsDetector.cpp
thattil: Sounds great! Changed it to the popcounter :D

slsDetectorSoftware/include/Detector.h
thattil: done

slsDetectorServers/mythen3DetectorServer/slsDetectorFunctionList.c
thattil: done. works!

thattil:
Fixed it in API and included popcount.
Most of these tests will fail for mythen3 and gotthard2 until we have at least the basic requirements fulfilled. They are still work in progress. The tests for mythen3 are counters, dacs, clkdiv, clkphase, clkdivider and maxclkphaseshift. They work.

PR #76, opened 2020-02-03
Buffern in exec

erikfrojdh:
Was there a specific reason for having the try catch block? I don't think the c functions should throw...
Also removed the vlarray which is non standard and replaced with a fixed size buffer.


thattil:
true

PR #79, opened 2020-02-25
Bursts

thattil:


slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
```diff
@@ -748,12 +758,39 @@ int64_t getNumFrames() {
 void setNumTriggers(int64_t val) {
     if (val > 0) {
 		FILE_LOG(logINFO, ("Setting number of triggers %lld\n", (long long int)val));
```
erikfrojdh: I would say that the cast is not needed

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
```diff
@@ -748,12 +758,39 @@ int64_t getNumFrames() {
...
+}
+
+int64_t getNumBursts() {
+	if (burstMode != BURST_OFF && getTiming() == AUTO_TIMING) {
+		return get64BitReg(SET_CYCLES_LSB_REG, SET_CYCLES_MSB_REG);
+	}
+	return numBursts;
 }
 
 int setExpTime(int64_t val) {
```
erikfrojdh: Not displayed here but looks like setExptimeBurst and setExptimeCont calls the same function. Maybe I misunderstood the logic but seems confusing.

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -1253,6 +1279,29 @@ void slsDetector::setNumberOfTriggers(int64_t value) {
     shm()->nTriggers = value;
     sendTotalNumFramestoReceiver();
 }
+
+int64_t slsDetector::getNumberOfBurstsFromShm() {
```
erikfrojdh: Is this function used? When searching I don't find it...

slsDetectorSoftware/src/slsDetector.cpp
```diff
@@ -1253,6 +1279,29 @@ void slsDetector::setNumberOfTriggers(int64_t value) {
...
+    return shm()->nBursts; 
+}
+
+int64_t slsDetector::getNumberOfBursts() {
+    int64_t prevVal = shm()->nBursts;
+    int64_t retval = -1;
+    sendToDetector(F_GET_NUM_BURSTS, nullptr, retval);
+    FILE_LOG(logDEBUG1) << "number of bursts :" << retval;
+    shm()->nBursts = retval;
+    if (prevVal != retval) {
```
erikfrojdh: by comparing against shm()->nBursts you can skip using prevVal

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
thattil: Yes. This is for now. Soon, the plan is to have different registers.

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
thattil: done

slsDetectorSoftware/src/slsDetector.cpp
thattil: I missed the progress by multi. Removed all the getNumberofFramesfromShm, triggersfromshm, and burstsfromshm.. and made into one function to get getTotalNumFramestoReceive.

slsDetectorSoftware/src/slsDetector.cpp
thattil: but i need to update shm()->nBursts before calling sendTotalNumFrametoReceiver().
If I do it before the if condition, I need a new variable.
If I do it inside the if condition, I need to repeat the asignment in the else.

thattil: done

thattil:
The nbursts is needed to calculate progress in multi. It is used instead of triggers in gotthard2 for auto and burst mode.

PR #82, opened 2020-03-03
Moench updated

thattil:
- integrated moench server with new firmware (rewrite of moench server)
- new settings list for moench
- removed moench from gain plot
- moench nx and ny are calculated from adcenable mask and #samples (not hardcoded anymore)
- updated moench dacs with proper names (gui as well)
- code to read default pattern text in server
- detangling chiptestboard and moench functions (especially Detector API and command line)
- moench (analog only, #samples must be divisible by 2; adcmask must have matching top and bottom adc masks)
-shm()->nchanx, nchany, asamples etc can be removed in  another pr for other detectors as well


erikfrojdh:
Looks ok to me. Follows the previous style. Later we might have the opportunity to refactor on the server side.

PR #83, opened 2020-03-04
New logger

erikfrojdh:
Small cleanup of the logger class. Removed a buch of template and macro stuff that we don't use. 

The old one is still there in logger.h and can be used with FILE_LOG
The new one is in logger2.h and can be used with LOG

At some point where we can swap. Compiles on RH7


erikfrojdh:
Added std::flush to the output. It dosen't seem to affect performance

PR #86, opened 2020-03-09
ctb, moench, eiger, gotthard: get number of channels for moench and c…

thattil:
…tb, others removed unnecessary variables in shm, added moench virtual sever


erikfrojdh:
Looks good! Code is more straight forward now. I'll merge it!

PR #89, opened 2020-03-26
Gappixels

thattil:
gap pixels only for call back
gap pixels for jungfrau
removed gap pixels from shm
json add header to map
zmqHeader instead of many arguments
completeImage added to zmqHeader
virtual servers (txndelay, imagetest, eiger data (fix 32 bit, 1g data), removed warnings, )
removed rxr cache or send update to receiver


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -1612,24 +1612,27 @@ void Detector::setPatternBitMask(uint64_t mask, Positions pos) {
...
 }
 
 Result<std::string> Detector::getAdditionalJsonParameter(const std::string &key,
                                                          Positions pos) const {
     return pimpl->Parallel(&Module::getAdditionalJsonParameter, pos, key);
 }
 
-void Detector::setAdditionalJsonParameter(const std::string &key,
-                                          const std::string &value,
+void Detector::setAdditionalJsonParameter(const std::map<std::string, std::string> para,
```
erikfrojdh: Do we need a map here, isn't it enough to take const std::string& key, const std::string& parameter? If we just set one parameter?

slsDetectorSoftware/src/Module.h
```diff
@@ -988,36 +979,14 @@ class Module : public virtual slsDetectorDefs {
...
-     * append it
-     * @param key additional json header parameter
-     * @param value additional json header parameter value (cannot be empty)
-     * @returns the additional json header parameter value,
-     * empty if no parameter found in additional json header
-     */
-    std::string setAdditionalJsonParameter(const std::string &key,
-                                           const std::string &value);
+    /** empty vector deletes entire additional json header */
+    void setAdditionalJsonHeader(const std::map<std::string, std::string> jsonHeader);
```
erikfrojdh: Since we use map as const we could probably use a reference

slsReceiverSoftware/src/ClientInterface.cpp
```diff
@@ -1427,4 +1405,23 @@ int ClientInterface::increment_file_index(Interface &socket) {
         impl()->setFileIndex(impl()->getFileIndex() + 1);
     }
     return socket.Send(OK);
-}
\ No newline at end of file
+}
+
+
+int ClientInterface::set_additional_json_parameter(Interface &socket) {
```
erikfrojdh: Do we need special function for setting and getting? Couldn't we just put both key and value in one char array?

thattil: I didnt really get you. The get function for json parameter is not special here as it uses only one char array. The set needs 2d char array. Using 1 d char array to include 2 strings is a bit ugly I feel, and also does not achieve any particular purpose. copyDetectorServer also uses 2 char arrays. Or did I misunderstand what you meant?

slsDetectorSoftware/src/Module.h
thattil: Yes, will fix this

slsDetectorSoftware/src/Detector.cpp
thattil: I agree. It came from wanting to allow multiple json parameters, but removing this later. Will fix this.

slsDetectorSoftware/src/Module.h
thattil: done

PR #96, opened 2020-05-11
More receiver functions throws 

erikfrojdh:
Cleaning up Module.cpp

Mode receiver specific function throws if the receiver is not configured. 
Throw moved to sendToReceiver (previous)

### DIFF
--- a/slsDetectorSoftware/src/DetectorImpl.cpp
+++ b/slsDetectorSoftware/src/DetectorImpl.cpp
@@ -1048,7 +1048,8 @@ int DetectorImpl::acquire() {
             }
             Parallel(&Module::startAndReadAll, {});
         } catch (...) {
-            Parallel(&Module::stopReceiver, {});
+            if (receiver)
+                Parallel(&Module::stopReceiver, {});
             throw;
         }
 
--- a/slsDetectorSoftware/src/Module.cpp
+++ b/slsDetectorSoftware/src/Module.cpp
@@ -625,11 +625,7 @@ int Module::getControlPort() const { return shm()->controlPort; }
 int Module::getStopPort() const { return shm()->stopPort; }
 
 bool Module::lockServer(int lock) {
-    int retval = -1;
-    LOG(logDEBUG1) << "Setting detector server lock to " << lock;
-    sendToDetector(F_LOCK_SERVER, lock, retval);
-    LOG(logDEBUG1) << "Lock: " << retval;
-    return retval == 1;
+    return sendToDetector<int>(F_LOCK_SERVER, lock) != 0;
 }
 
 sls::IpAddr Module::getLastClientIP() {
@@ -724,11 +720,8 @@ std::vector<std::string> Module::getConfigFileCommands() {
 }
 
 slsDetectorDefs::detectorSettings Module::getSettings() {
-    int arg = -1;
-    int retval = -1;
-    sendToDetector(F_SET_SETTINGS, arg, retval);
-    LOG(logDEBUG1) << "Settings: " << retval;
-    return static_cast<detectorSettings>(retval);
+    auto r = sendToDetector<int>(F_SET_SETTINGS, -1);
+    return static_cast<detectorSettings>(r);
 }
 
 void Module::setSettings(detectorSettings isettings) {
@@ -758,16 +751,10 @@ int Module::getThresholdEnergy() {
             return -1;
         }
     }
-
-    LOG(logDEBUG1) << "Getting threshold energy";
-    int retval = -1;
-    sendToDetector(F_GET_THRESHOLD_ENERGY, nullptr, retval);
-    LOG(logDEBUG1) << "Threshold: " << retval;
-    return retval;
+    return sendToDetector<int>(F_GET_THRESHOLD_ENERGY);
 }
 
 void Module::setThresholdEnergy(int e_eV, detectorSettings isettings, int tb) {
-
     // check as there is client processing
     if (shm()->myDetectorType == EIGER) {
         setThresholdEnergyAndSettings(e_eV, isettings, tb);
@@ -1208,11 +1195,7 @@ int64_t Module::getMeasurementTime() const {
 }
 
 slsDetectorDefs::timingMode Module::getTimingMode() {
-    int arg = -1;
-    timingMode retval = GET_TIMING_MODE;
-    sendToDetector(F_SET_TIMING_MODE, arg, retval);
-    LOG(logDEBUG1) << "Timing Mode: " << retval;
-    return retval;
+    return sendToDetector<timingMode>(F_SET_TIMING_MODE, -1);
 }
 
 void Module::setTimingMode(timingMode value) {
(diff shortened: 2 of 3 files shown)


slsDetectorSoftware/src/Module.cpp
```diff
@@ -1722,10 +1700,7 @@ void Module::updateReceiverStreamingIP() {
 }
 
 bool Module::getTenGigaFlowControl() {
-    int retval = -1;
-    sendToDetector(F_GET_TEN_GIGA_FLOW_CONTROL, nullptr, retval);
-    LOG(logDEBUG1) << "ten giga flow control :" << retval;
-    return retval == 1;
+    return sendToDetector<int>(F_GET_TEN_GIGA_FLOW_CONTROL)!=0;
```
thattil, line 1703: I think we can return an int, and it will cast itself into bool, right? We dont need to check if its not equal to 0

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2066,29 +2032,22 @@ void Module::setBurstMode(slsDetectorDefs::burstMode value) {
...
-    LOG(logDEBUG1) << "Current source enable:" << retval;
-    return static_cast<bool>(retval);
+    return sendToDetector<int>(F_GET_CURRENT_SOURCE);
 }
 
 void Module::setCurrentSource(bool value) {
-    int arg = static_cast<int>(value);
-    LOG(logDEBUG1) << "Setting current source enable to " << arg;
-    sendToDetector(F_SET_CURRENT_SOURCE, arg, nullptr);
+    LOG(logDEBUG1) << "Setting current source enable to " << value;
```
thattil, line 2039: Do we need this printout? since you already have it in sendtoDetector right?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2066,29 +2032,22 @@ void Module::setBurstMode(slsDetectorDefs::burstMode value) {
...
-    return static_cast<slsDetectorDefs::timingSourceType>(retval);
+    auto r = sendToDetector<int>(F_GET_TIMING_SOURCE);
+    return static_cast<slsDetectorDefs::timingSourceType>(r);
 }
 
 void Module::setTimingSource(slsDetectorDefs::timingSourceType value) {
-    int arg = static_cast<int>(value);
-    LOG(logDEBUG1) << "Setting timing source to " << arg;
-    sendToDetector(F_SET_TIMING_SOURCE, arg, nullptr);
+    LOG(logDEBUG1) << "Setting timing source to " << value;
```
thattil, line 2049: same

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2105,22 +2064,16 @@ void Module::setROI(slsDetectorDefs::ROI arg) {
         arg.xmin = -1;
         arg.xmax = -1;
     }
-    std::array<int, 2> args{arg.xmin, arg.xmax};
     LOG(logDEBUG) << "Sending ROI to detector [" << arg.xmin << ", " << arg.xmax
                   << "]";
-    sendToDetector(F_SET_ROI, args, nullptr);
+    sendToDetector(F_SET_ROI, arg, nullptr);
```
thattil, line 2069: can we send the ROI structure to the detector? being 64 and 32 bit is okay?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2609,41 +2536,16 @@ std::string Module::printReceiverConfiguration() {
...
-void Module::execReceiverCommand(const std::string &cmd) {
-    char arg[MAX_STR_LENGTH]{};
-    char retval[MAX_STR_LENGTH]{};
-    sls::strcpy_safe(arg, cmd.c_str());
-    LOG(logDEBUG1) << "Sending command to receiver: " << arg;
-    if (shm()->useReceiverFlag) {
-        sendToReceiver(F_EXEC_RECEIVER_COMMAND, arg, retval);
-        LOG(logINFO) << "Receiver " << detId << " returned:\n" << retval;
-    }
+    sendToReceiver(F_EXIT_RECEIVER, nullptr, nullptr);
```
thattil, line 2548: can we not add sendToReceiver with only fnum, like you did for sendtodetector?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2725,40 +2627,25 @@ void Module::setPartialFramesPadding(bool padding) {
...
-        LOG(logDEBUG1) << "Frames Caught by Receiver: " << retval;
-    }
-    return retval;
+    return sendToReceiver<int64_t>(F_GET_RECEIVER_FRAMES_CAUGHT);
 }
 
 std::vector<uint64_t> Module::getNumMissingPackets() const {
+    //TODO!(Erik) Refactor
     LOG(logDEBUG1) << "Getting num missing packets";
     if (shm()->useReceiverFlag) {
```
thattil, line 2650: You can also remove the check for useReceiverflag like you did for the other 'onlyreceiver' functions..

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1048,7 +1048,8 @@ int DetectorImpl::acquire() {
             }
             Parallel(&Module::startAndReadAll, {});
         } catch (...) {
-            Parallel(&Module::stopReceiver, {});
+            if (receiver)
+                Parallel(&Module::stopReceiver, {});
```
thattil, line 1052: Good catch!

slsDetectorSoftware/src/Module.cpp
erikfrojdh: Done

slsDetectorSoftware/src/Module.cpp
erikfrojdh: done

slsDetectorSoftware/src/Module.cpp
erikfrojdh: done

slsDetectorSoftware/src/Module.cpp
erikfrojdh: Struct is declared with attribute packed, should be the same as sending two ints.

slsDetectorSoftware/src/Module.cpp
erikfrojdh: Depends if we want to be able to do sendToReceiver<int>(fnum), we cant have both... since the only difference in signature is the return value overload resolution fails.

slsDetectorSoftware/src/Module.cpp
erikfrojdh: The check is in, sendToReceiver and here we use the socket directly. Is there a reason we cant use sendToReceiver here?

slsDetectorSoftware/src/DetectorImpl.cpp
erikfrojdh: Danke!

slsDetectorSoftware/src/Module.cpp
thattil: Alright! :+1:

slsDetectorSoftware/src/Module.cpp
thattil: ahh okay, then never mind, its too few in this case.

slsDetectorSoftware/src/Module.cpp
thattil: I guess knowing detector type and number of interfaces should help to know how many ports. Then we could stick to sendToReceiver (also in clientInterface). If you like, I can fix that after you push.

PR #97, opened 2020-05-11
Eiger26

thattil:
1. set master, top (hardware, 0, 1) to Beb and Feb. Hardware is if no top/master set in config file.
2. Read master, top only from Beb from Beb_getmoduleconfiguration.
3. Config file read only at end of set up
4. stop server starts after a wait (waiting for control server to setup top/master)
5. virtual servers read config file (not get moduleconfiguration)
6. When setting to hardware (no top/master in config), it has to also set top in Beb and master in Feb for other usages (detector position, configuremac etc).
7. Deactivated at end of setup, activated when setting hostname
8. Detector Id changes in fw
9. Quad (only sets top right fpga)


slsDetectorServers/eigerDetectorServer/Beb.c
```diff
@@ -298,125 +297,146 @@ int Beb_IsTransmitting(int *retval, int tengiga, int waitForDelay) {
...
 
-    return ret;
+    Beb_close(fd, csp0base);
+    return 1;
 }
 
-int Beb_Activate(int enable) {
-    // mapping new memory
+int Beb_SetActivate(int enable) {
+    enable = enable == 0 ? 0 : 1;
```
erikfrojdh, line 394: I had to think a bit before I understood what and why =)
I would move the statement to just before assigning to Beb_activated an maybe replace either with the 
short but cryptical:

`Beb_activated = !!enable;`

or the more readable:
`if (enable)
   enable = 1;
Beb_activated = enable;`

erikfrojdh: Not one liners... but the editor doesn't like me =)

slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1681,16 +1875,39 @@ int getBebFPGATemp() {
...
 #else
-    int ret = Beb_Activate(enable);
-    Feb_Control_activate(ret);
-    return ret;
+    if (!Beb_SetActivate(enable)) {
+        return FAIL;
+    }
+    Feb_Control_activate(enable);
 #endif
+    if (enable == 0) {
```
erikfrojdh, line 1892: I would switch the  statement around and do if(enabled)

thattil: done

slsDetectorServers/eigerDetectorServer/Beb.c
thattil: I did a if (enable < 0 ) print error and return 0. Otherwise, we dont care if enable > 1. I removed the line enable = enable == 0... from Beb.c and list.c

PR #100, opened 2020-05-20
Mythen3

thattil:


slsDetectorSoftware/src/Module.cpp
```diff
@@ -1036,24 +1027,64 @@ void Module::setNumberOfDigitalSamples(int value) {
...
+                   << "ns (gateindex: " << gateIndex << ")";
+    int64_t args[]{static_cast<int64_t>(gateIndex), value};
+    sendToDetector(F_SET_EXPTIME, args, nullptr);
     if (shm()->useReceiverFlag) {
         LOG(logDEBUG1) << "Sending exptime to Receiver: " << value;
-        sendToReceiver(F_RECEIVER_SET_EXPTIME, value, nullptr);
+        sendToReceiver(F_RECEIVER_SET_EXPTIME, args, nullptr);
     }
     if (prevVal != value) {
         updateRateCorrection();
```
erikfrojdh, line 1059: Not for this PR but an idea, could this check be done in the server? Meaning that every time we set exptime the ratecorr gets updated. (if enabled) and the server tracks to see if it's a new exptime or not. Then we need less logic in the client

slsDetectorSoftware/src/Module.cpp
```diff
@@ -1281,10 +1312,15 @@ int Module::getADC(dacIndex index) {
...
-    LOG(logDEBUG1) << "Setting signal flag to " << pol;
+Module::getExternalSignalFlags(int signalIndex) {
     return sendToDetector<slsDetectorDefs::externalSignalFlag>(
-        F_SET_EXTERNAL_SIGNAL_FLAG, pol);
+        F_GET_EXTERNAL_SIGNAL_FLAG, signalIndex);
+}
+
+void Module::setExternalSignalFlags(int signalIndex, externalSignalFlag type) {
+    LOG(logDEBUG1) << "Setting signal flag (" << signalIndex << ") to " << type;
+    int args[2] = {signalIndex, static_cast<int>(type)};
```
erikfrojdh, line 1322: int args[]{signalIndex,...} should work. 
or std::array<int, 2>args{signalIn..}

slsReceiverSoftware/src/BinaryFile.cpp
```diff
@@ -191,6 +191,13 @@ void BinaryFile::CreateMasterFile(bool masterFileWriteEnable,
                 "Dbit Offset                : %d\n"
                 "Dbit Bitset                : %lld\n"
                 "Roi (xmin, xmax)           : %d %d\n"
+                "Exptime1 (ns)              : %lld\n"
```
erikfrojdh, line 194: I guess we want to keep the master file the same for all detectors? But now we write many fields that does not exists for most detectors

slsDetectorGui/CMakeLists.txt
```diff
@@ -89,6 +89,7 @@ target_include_directories(slsDetectorGui PUBLIC
 
 target_link_libraries(slsDetectorGui PUBLIC
     slsDetectorShared
+    slsSupportLib
```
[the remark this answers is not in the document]
thattil: I added this line for logger.h. If its not needed, I can remove it. I can add The slsProjectWarnings.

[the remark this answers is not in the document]
thattil: does it also need slsProjectOptions?

slsDetectorSoftware/src/Module.cpp
thattil: This could, but I wanted it to be also changed in the receiver, before doing updateratecorrection. If there is an error in ratecorrection, which almost certainly could be (if dr is not 16 or 32), then the receiver does not get updated because of the exception. This is why I put it here in the client, after all the exptime is actually set for detector and receiver.

slsDetectorSoftware/src/Module.cpp
thattil: done

slsReceiverSoftware/src/BinaryFile.cpp
thattil: Yes, this is something I would like to change in the future.

slsDetectorGui/CMakeLists.txt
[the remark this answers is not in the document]
erikfrojdh: No slsProjectOptions is public so that should automatically be included

PR #101, opened 2020-05-25
Vetoheader

thattil:
gotthard2 veto header not standard


slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
```diff
@@ -2274,7 +2274,7 @@ void *start_timer(void *arg) {
     int datasize = imagesize;
     int packetsize = datasize + sizeof(sls_detector_header);
     int vetodatasize = VETO_DATA_SIZE;
-    int vetopacketsize = vetodatasize + sizeof(sls_detector_header);
+    int vetopacketsize = vetodatasize + VETO_HEADER_SIZE;
```
erikfrojdh, line 2277: Why not use sizeof(veto_header)?

slsReceiverSoftware/src/GeneralData.h
```diff
@@ -490,6 +497,7 @@ class Gotthard2Data : public GeneralData {
         vetoDataSize = 160;
         vetoPacketSize = headerSizeinPacket + vetoDataSize;
         vetoImageSize = vetoDataSize * packetsPerFrame;
+        vetoHsize = 24;
```
erikfrojdh, line 500: Same here don't we have the header somewhere so that we could use sizeof?

thattil: No, its in the sever definitions.  Not accsessible here.

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
thattil: Yes.will update

PR #102, opened 2020-06-05
exposing receiver thread ids to client

thattil:


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -141,6 +141,55 @@ void Detector::setDetectorSize(const defs::xy value) {
     pimpl->setNumberOfChannels(value);
 }
 
+std::vector<defs::detectorSettings> Detector::getSettingsList() const {
```
erikfrojdh, line 144: Sine each settings vector is unique for the detector we don't need to build it up using push_back. The vector can be constructed with the values needed:
```
case defs::EIGER:
     return std::vector<defs::detectorSettings>{defs::STANDARD, defs::HIGHGAIN ....}
```

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -405,6 +454,104 @@ Result<int> Detector::getTemperature(defs::dacIndex index,
     return res;
 }
 
+std::vector<defs::dacIndex> Detector::getDacList() const {
+    std::vector<defs::dacIndex> retval;
+    int i = 0;
```
erikfrojdh, line 459: I is not used until the for loop so it can be moved into the loop declaration

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -405,6 +454,104 @@ Result<int> Detector::getTemperature(defs::dacIndex index,
...
+    switch (getDetectorType().squash()) {
+    case defs::EIGER:
```
erikfrojdh, line 461: Same as for settings, we don't really need to build up the vector. 
Probably also workes with just return {dac, dac, dac, ...}

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2614,6 +2614,11 @@ sls::IpAddr Module::getReceiverLastClientIP() const {
     return sendToReceiver<sls::IpAddr>(F_GET_LAST_RECEIVER_CLIENT_IP);
 }
 
+std::array<pid_t, NUM_RX_THREAD_IDS> Module::getReceiverThreadIds() const {
```
erikfrojdh, line 2617: It would be possible to return an sls::StaticVector if you only want to print the active threads. But it's fine to keep zeros as well

slsDetectorSoftware/src/CmdProxy.cpp
```diff
@@ -893,22 +925,26 @@ std::string CmdProxy::DacValues(int action) {
         os << "\n\tGets the list of commands for every dac for this detector."
            << '\n';
     } else if (action == defs::GET_ACTION) {
-        std::vector<std::string> names = DacCommands();
-        std::vector<std::string> res(names.size());
-        std::vector<std::string> args;
-        for (size_t i = 0; i < names.size(); ++i) {
+        auto det_type = det->getDetectorType().squash(defs::GENERIC);
+        auto t = det->getDacList();
+        std::vector<std::string> res(t.size());
```
erikfrojdh, line 930: Works but would be more elegant to to have to call through the command line interface. But I guess the CTB is the problem here.

slsDetectorSoftware/src/Detector.cpp
thattil: done

slsDetectorSoftware/src/Detector.cpp
thattil: done

slsDetectorSoftware/src/CmdProxy.cpp
thattil: Yes, because if we use ToString for ctb as well, we will need to give the detector type as argument as well

slsDetectorSoftware/src/Detector.cpp
thattil: done

PR #117, opened 2020-07-22
Readlink

thattil:


erikfrojdh:
Looks good, but the function for getting the path is repeated in many places. Is is possible to make a function that would return the path? 

then it could work something like this
`char fpath[128];
getabspath(fpath, CONFIG_FILE);
fopen...`

thattil:
> Looks good, but the function for getting the path is repeated in many places. Is is possible to make a function that would return the path?
> 
> then it could work something like this
> 
> ```c
> char fpath[128];
> getabspath(fpath, CONFIG_FILE);
> fopen...
> ```

done! moved to common.c

PR #121, opened 2020-07-24
Check args for sending to detector/receiver

erikfrojdh:
- Reduced code duplication between const and non const sendToDetector/sendToReceiver
- Static asserts to prevent from passing in pointers to the templated functions
- Runtime check that the combination of size and nullptr is sane

### DIFF
--- a/slsDetectorSoftware/src/Module.cpp
+++ b/slsDetectorSoftware/src/Module.cpp
@@ -58,7 +58,7 @@ void Module::freeSharedMemory() {
     }
 }
 
-bool Module::isFixedPatternSharedMemoryCompatible() {
+bool Module::isFixedPatternSharedMemoryCompatible() const {
     return (shm()->shmversion >= SLS_SHMAPIVERSION);
 }
 
@@ -84,15 +84,15 @@ void Module::setHostname(const std::string &hostname,
     }
 }
 
-int64_t Module::getFirmwareVersion() {
+int64_t Module::getFirmwareVersion() const {
     return sendToDetector<int64_t>(F_GET_FIRMWARE_VERSION);
 }
 
-int64_t Module::getDetectorServerVersion() {
+int64_t Module::getDetectorServerVersion() const {
     return sendToDetector<int64_t>(F_GET_SERVER_VERSION);
 }
 
-int64_t Module::getSerialNumber() {
+int64_t Module::getSerialNumber() const {
     return sendToDetector<int64_t>(F_GET_SERIAL_NUMBER);
 }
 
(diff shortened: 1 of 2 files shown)


slsDetectorSoftware/src/Module.cpp
```diff
@@ -2386,93 +2386,158 @@ uint64_t Module::getReceiverCurrentFrameIndex() const {
 
 // private
 
+void Module::preSendArgsCheck(const void *args, size_t args_size, void *retval,
+                              size_t retval_size) const {
```
thattil, line 2390: just a sendArgsCheck?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2386,93 +2386,158 @@ uint64_t Module::getReceiverCurrentFrameIndex() const {
...
 
+template <typename Arg>
+void Module::sendToDetector(int fnum, const Arg &args, std::nullptr_t) {
+    static_cast<const Module &>(*this).sendToDetector(fnum, args, nullptr);
+}
+
 template <typename Ret>
-void Module::sendToDetector(int fnum, std::nullptr_t, Ret &retval) {
+void Module::sendToDetector(int fnum, std::nullptr_t, Ret &retval) const {
     LOG(logDEBUG1) << "Sending: ["
```
thattil, line 2464: Since now we have const sendToDetector, we can set many of the get Module class functions to a const..

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2386,93 +2386,158 @@ uint64_t Module::getReceiverCurrentFrameIndex() const {
...
+    // the return type
+    preSendArgsCheck(args, args_size, retval, retval_size);
     auto stop = DetectorSocket(shm()->hostname, shm()->stopPort);
     stop.sendCommandThenRead(fnum, args, args_size, retval, retval_size);
     stop.close();
 }
 
-template <typename Arg, typename Ret>
-void Module::sendToDetectorStop(int fnum, const Arg &args, Ret &retval) {
-    LOG(logDEBUG1) << "Sending to Stop: ["
```
thattil, line 2470: Why are you removing these prints? Are they moved elsewhere? They are removed in a few other places as well..

erikfrojdh: The prints are in the const version of the function, and the non const version just calls the const version. So all prints should still be there, just not duplicated for const and non const. 

Idea is taken from Scott Meyers Effective C++... This way of casting can be used when non const version is not really modifying anything, just that it's called from a non const context.

slsDetectorSoftware/src/Module.cpp
erikfrojdh: Indeed, I put this off before since I didn't like the code duplication. Should I add it in this PR?

slsDetectorSoftware/src/Module.cpp
erikfrojdh: checkArgs?

thattil: sure

slsDetectorSoftware/src/Module.cpp
thattil: upto you

slsDetectorSoftware/src/Module.cpp
thattil: soundd good :)

PR #122, opened 2020-07-29
G2

thattil:
gotthard2
-  vetophoton now has gain thresholds for each row and the get writes to file gain indices and adu for each channel
- adus written in dec instead of hex
- veto file in server removed and uses vetophoton instead


slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -6329,35 +6328,37 @@ int get_inject_channel(int file_des) {
...
-    int args[3] = {-1, -1, -1};
 
+    int args[2] = {-1, -1};
     if (receiveData(file_des, args, sizeof(args), INT32) < 0)
         return printSocketReadError();
-    int values[args[2]];
+    int chipIndex = args[0];
+    int numChannels = args[1];
+
+    int gainIndices[args[1]];
```
erikfrojdh, line 6338: How large is the gainIndicies? I'm generally not so keen on VLA... We don't know the size of this before? Or if it's rather large maybe heap allocation is better?

slsDetectorSoftware/include/Detector.h
```diff
@@ -980,9 +980,9 @@ class Detector {
     void setInjectChannel(const int offsetChannel, const int incrementChannel,
                           Positions pos = {});
 
-    /** [Gotthard2] adu values for each channel */
-    Result<std::vector<int>> getVetoPhoton(const int chipIndex,
-                                           Positions pos = {});
+    /** [Gotthard2] gain indices and adu values for each channel */
+    void getVetoPhoton(const int chipIndex, const std::string &fname,
```
erikfrojdh, line 984: I wonder if we should look through the API and rename functions that read or write files to make them stand out from the normal set/get?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -1395,14 +1427,25 @@ std::vector<int> Module::getVetoPhoton(const int chipIndex) const {
...
     client.Receive(&nch, sizeof(nch));
-
-    int adus[nch];
-    memset(adus, 0, sizeof(adus));
-    client.Receive(adus, sizeof(adus));
-    std::vector<int> retvals(adus, adus + nch);
     LOG(logDEBUG1) << "Getting veto photon [" << chipIndex << "]: " << nch
                    << " channels\n";
-    return retvals;
+    int gainIndices[nch];
```
erikfrojdh, line 1432: While VLAs and memset might be the way to do things in C I would prefer not to use it in C++. Could you maybe read into a vector?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -1429,14 +1472,15 @@ void Module::setVetoPhoton(const int chipIndex, const int numPhotons,
...
+    LOG(logDEBUG1) << "Setting veto photon. Reading Gain values from file";
 
     int totalEnergy = numPhotons * energy;
     int ch = shm()->nChan.x;
-    int gainIndex = 2;
     int nRead = 0;
-    int value[ch];
-    memset(value, 0, sizeof(value));
-    bool firstLine = true;
+    int values[ch];
```
erikfrojdh, line 1480: same as above? Vector?

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
```diff
@@ -387,6 +388,7 @@ void setupDetector() {
     for (int i = 0; i < NCHIP; ++i) {
         for (int j = 0; j < NCHAN; ++j) {
             vetoReference[i][j] = 0;
+            vetoGainIndices[i][j] = 0;
```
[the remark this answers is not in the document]
thattil: done

slsDetectorSoftware/include/Detector.h
thattil: like setVetoPhotonFile? and getVetoPhotonFile? In this case, the file uploaded and the file downloaded are different.. but yes sure.. maybe in a different PR though..

slsDetectorSoftware/src/Module.cpp
thattil: done

slsDetectorSoftware/src/Module.cpp
thattil: done

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
thattil: Its not large, and the size is not known at compilation (numchannels depends on detector). More for extra validation between client and server constants. Adding const in this case to not make it VLA.

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
```diff
@@ -1786,56 +1788,58 @@ void getInjectedChannels(int *offset, int *increment) {
 int setVetoReference(int gainIndex, int value) {
     LOG(logINFO, ("Setting veto reference [chip:-1, G%d, value:0x%x]\n",
                   gainIndex, value));
-    int vals[NCHAN];
-    memset(vals, 0, sizeof(vals));
+    int values[NCHAN];
+    memset(values, 0, sizeof(values));
+    int gainIndices[NCHAN];
+    memset(gainIndices, 0, sizeof(gainIndices));
```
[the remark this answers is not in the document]
thattil: done

slsDetectorSoftware/include/Detector.h
erikfrojdh: Makes sense!

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -6332,16 +6332,14 @@ int set_veto_photon(int file_des) {
...
 
-    int gainIndices[args[1]];
-    memset(gainIndices, 0, sizeof(gainIndices));
+    int gainIndices[numChannels];
     if (receiveData(file_des, gainIndices, sizeof(gainIndices), INT32) < 0)
         return printSocketReadError();
 
-    int values[args[1]];
-    memset(values, 0, sizeof(values));
+    int values[numChannels];
```
erikfrojdh, line 6342: In this case you will still get a VLA, since numChannels is not known at compile time. 
It's only in the case below that you avoid a VLA
```cpp
const int size = 5; //this is known at compile time
int arr[size]; // now the array size is fixed
```

But since VLAs are used a quite a bit in the servers we don't need to address all the cases in this PR.

slsDetectorSoftware/src/Module.cpp
```diff
@@ -1389,19 +1389,25 @@ void Module::setInjectChannel(const int offsetChannel,
     sendToDetector(F_SET_INJECT_CHANNEL, args, nullptr);
 }
 
-void Module::sendVetoPhoton(const int chipIndex, int *gainIndices,
-                            int *values) {
+void Module::sendVetoPhoton(const int chipIndex, std::vector<int> gainIndices,
```
erikfrojdh, line 1392: Can use a const& to the vector in order to avoid copying the data.

slsDetectorSoftware/src/Module.cpp
```diff
@@ -1389,19 +1389,25 @@ void Module::setInjectChannel(const int offsetChannel,
...
-    int nch = shm()->nChan.x;
     int args[]{chipIndex, nch};
-    LOG(logDEBUG1) << "Sending veto photon/file to detector [chip:" << chipIndex
-                   << ", nch:" << nch << "]";
     auto client = DetectorSocket(shm()->hostname, shm()->controlPort);
     client.Send(&fnum, sizeof(fnum));
     client.Send(args, sizeof(args));
-    client.Send(gainIndices, sizeof(int) * nch);
-    client.Send(values, sizeof(int) * nch);
+    client.Send(gainIndices.data(), sizeof(int) * nch);
```
erikfrojdh, line 1409: In case we ever would change the data of the vector this is slightly less error prone:

```cpp
client.Send(gainIndices.data(), sizeof(decltype(gainIndicies::value_type)) * nch);

```

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
thattil: alright. also because we have to use VLAs in the server as we dont know the size for many of them at compile time.

slsDetectorSoftware/src/Module.cpp
thattil: done

slsDetectorSoftware/src/Module.cpp
thattil: done, sizeof(decltype(values[0])) * nch

PR #124, opened 2020-07-30
Remove VLAs 

erikfrojdh:
Since the Linux kernel managed to get rid of their use of VLAs maybe we can to =)


slsDetectorSoftware/src/Module.cpp
```diff
@@ -1679,27 +1678,23 @@ void Module::getBadChannels(const std::string &fname) const {
...
     // receive badchannels
     int nch = -1;
-    std::vector<int> badchannels;
     client.Receive(&nch, sizeof(nch));
+    std::vector<int> badchannels(nch);
     if (nch > 0) {
-        int temp[nch];
-        memset(temp, 0, sizeof(temp));
-        client.Receive(temp, sizeof(temp));
-        badchannels.insert(badchannels.end(), &temp[0], &temp[nch]);
```
thattil, line 1688: do you understand why this was inserting backwards? I think this might have been a bug that we had before?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -2139,30 +2133,31 @@ void Module::startPattern() { sendToDetector(F_START_PATTERN); }
...
-            client.Receive(retvals, sizeof(retvals));
-            for (int i = 0; i < size; ++i) {
-                retval[retvals[2 * i]] = retvals[2 * i + 1];
+            client.Receive(&buff[0], buff.size());
+            std::istringstream iss(buff);
+            std::string key, value;
+            while(iss >> key){
+                iss >> value;
+                retval[key] = value;
             }
```
thattil, line 2161: much  more readable! :)

slsReceiverSoftware/src/ClientInterface.cpp
```diff
@@ -901,14 +901,12 @@ int ClientInterface::get_frame_index(Interface &socket) {
...
-    uint64_t retval[retvalsize];
-    std::copy(std::begin(m), std::end(m), retval);
+    auto missing_packets = impl()->getNumMissingPackets();
+    LOG(logDEBUG1) << "missing packets:" << sls::ToString(missing_packets);
+    auto size = static_cast<int>(missing_packets.size());
     socket.Send(OK);
-    socket.Send(&retvalsize, sizeof(retvalsize));
-    socket.Send(retval, sizeof(retval));
+    socket.Send(size);
+    socket.Send(missing_packets.data(), sizeof(missing_packets[0])* missing_packets.size());
```
thattil, line 909: missing decltype? beore missing_packets[0]

slsDetectorSoftware/src/Module.cpp
```diff
@@ -1395,14 +1403,11 @@ std::vector<int> Module::getVetoPhoton(const int chipIndex) const {
     }
     int nch = -1;
     client.Receive(&nch, sizeof(nch));
-
-    int adus[nch];
-    memset(adus, 0, sizeof(adus));
-    client.Receive(adus, sizeof(adus));
-    std::vector<int> retvals(adus, adus + nch);
+    std::vector<int> adus(nch);
+    client.Receive(adus.data(), sizeof(adus[0]) * adus.size());
```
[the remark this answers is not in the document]
erikfrojdh: No should work without since we adus[0] gets the value. Works the same way for ints
```cpp
int i;
sizeof(i); // 4
sizeof(int) // 4
````

slsDetectorSoftware/src/Module.cpp
erikfrojdh: Empty vector, so inserting at the end() which is also the start.

slsDetectorSoftware/src/Module.cpp
erikfrojdh: Merci

slsReceiverSoftware/src/ClientInterface.cpp
erikfrojdh: Same as the above.

PR #125, opened 2020-08-03
Eiger

thattil:
- changed vhighvoltage to highvotlage
- gui: subframe index not shown (%d shown)
- roi structure used with constructor
- rate corrections for each rxr updated in rxr master file
- always set speed to full speed for dr 4, 8, 16
- detector dependant master file attributes made into a class with inheritance, with methods to write them to binary or hdf5 file
- added xy constructor and print in ToString
- timing mode in rx  master file


slsReceiverSoftware/src/MasterAttributes.h
```diff
@@ -0,0 +1,567 @@
...
+
+#include <chrono>
+using ns = std::chrono::nanoseconds;
+
+// versions
+#define HDF5_WRITER_VERSION   (6.1) // 1 decimal places
+#define BINARY_WRITER_VERSION (6.1) // 1 decimal places
+
+class MasterAttributes {
+  public:
```
erikfrojdh, line 22: If all members are public I would declare this as a struct instead of a class.

slsReceiverSoftware/src/MasterAttributes.h
```diff
@@ -0,0 +1,567 @@
...
+    slsDetectorDefs::detectorType detType{slsDetectorDefs::GENERIC};
+    slsDetectorDefs::timingMode timingMode{slsDetectorDefs::AUTO_TIMING};
+    uint32_t imageSize{0};
```
erikfrojdh, line 25: Do we want all the data fields to be declared in the base class or only the ones that are common for all detectors?

slsReceiverSoftware/src/MasterAttributes.h
```diff
@@ -0,0 +1,567 @@
...
+    ns exptime3{0};
+    ns gateDelay1{0};
+    ns gateDelay2{0};
+    ns gateDelay3{0};
+    uint32_t gates;
+
+    MasterAttributes(){};
+    virtual ~MasterAttributes(){};
+
+    virtual void WriteMasterBinaryAttributes(FILE *fd) {
```
erikfrojdh, line 55: By making this a pure virtual function we can at compile time protect against this since you can't create a base class object. Then we don't need the error message.

slsReceiverSoftware/src/MasterAttributes.h
```diff
@@ -0,0 +1,567 @@
...
+            << "Timing Mode                : " << sls::ToString(timingMode)
+            << '\n'
+            << "Image Size                 : " << imageSize << " bytes" << '\n'
+            << "Pixels                     : " << sls::ToString(nPixels) << '\n'
+            << "Max Frames Per File        : " << maxFramesPerFile << '\n'
+            << "Total Frames               : " << totalFrames << '\n';
+        return oss.str();
+    };
+
+    void WriteBinaryAttributes(FILE *fd, std::string message) {
```
erikfrojdh, line 76: Do we actually need to pass the file pointer? In theory we could just generate the string and have all writing done by BinaryFile/HDF5File... But maybe this is more tricky in the HDF5 case...

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -1105,10 +1109,27 @@ Result<ns> Detector::getRateCorrection(Positions pos) const {
...
+    updateRxRateCorrections();
+}
+
+void Detector::updateRxRateCorrections() {
+    // get tau from all modules and send to Rx index 0
+    if (getDetectorType().squash() == defs::EIGER) {
+        if (getUseReceiverFlag().squash(false)) {
+            // convert Result<ns> to std::vector<in64_t>
+            auto retval = getRateCorrection();
+            std::vector<int64_t> t(retval.size());
```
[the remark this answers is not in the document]
thattil: Ill do what you suggested first.. converting rxr to use chrono will have to be another PR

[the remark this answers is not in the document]

slsReceiverSoftware/src/MasterAttributes.h
thattil: ok I just renamed the class to struct.

slsReceiverSoftware/src/MasterAttributes.h
thattil: They cannot be called by the parent reference that we are using

slsReceiverSoftware/src/MasterAttributes.h
thattil: they are not implemented in the hdf5 part and so cant make it pure virtual

slsReceiverSoftware/src/MasterAttributes.h
thattil: yes, its complicated in the hdf5 case and henc to make it more consistent, i did the writign for both in this class.

slsReceiverSoftware/src/MasterAttributes.h
thattil: done

PR #129, opened 2020-08-05
Utility function to get a vector of set bits

erikfrojdh:
Template function to return a vector with the positions of all set bits in a variable.


slsDetectorSoftware/src/CmdProxy.cpp
```diff
@@ -1,4 +1,5 @@
 #include "CmdProxy.h"
+#include "bit_stuff.h"
```
thattil, line 2: you want to rename this from bit_stuff.h to maybe bit_manupulation.h or something similar?

PR #150, opened 2020-08-28
Gui slot call focus fix

thattil:
Problem: when tabbing, the fields get set, or when clicking on a field and outside, it gets set.

qlinedit.. kept editingFinished(), but using isModified() and setModified(false) to ensure its not called when using tab or focus out

qspinbox and qdoublespinbox, moved from editingFinished() to valueChanged() while disabled keyboard tracking in form


slsDetectorGui/src/qTabPlot.cpp
```diff
@@ -389,6 +389,48 @@ void qTabPlot::SetTitles() {
             SLOT(SetTitles()));
 }
 
+void qTabPlot::isXMinModified() {
```
thattil, line 392: These additional (in between) slots are necessary because all the chkXMin  (check box) and dispXMin and dispXMax (text fields) calls SetXRange... so had to to add IsXMinModified() when dispXMin editingFinished() gets called.. before calling SetXRange

PR #164, opened 2020-09-04
gui: adding 1D detectors by default in x direction, others in y dir

thattil:


erikfrojdh (commented):
Would it be possible to check detsize y == 1 instead of having a switch and a bool? This to me looks cleaner and has the advantage that it would work without updates if we add new detectors.

thattil:
> Would it be possible to check detsize y == 1 instead of having a switch and a bool? This to me looks cleaner and has the advantage that it would work without updates if we add new detectors.

Makes sense. done.

PR #166, opened 2020-09-07
Rxrmetadata

thattil:
- additional json add header and scan parameter added to rxr master file
- fixed updates in rxr that needed to go in rxParameters structure
- added onto master file other detector metadata in rxr
- additional json add header made to allow changes on the fly (locking using a boolean to prevent lock each time it streams a frame)


erikfrojdh (commented):
Looks good to me. Only concern is that if you change the json add header during the measurement you don't know from the master file that that header has been used for all frames

thattil:
> Looks good to me. Only concern is that if you change the json add header during the measurement you don't know from the master file that that header has been used for all frames

I had talked with anna (@anberga). The master file is in any case only created at the beginning of an acquisition (at startReceiver()). So, whatever she changes during the acquisition (she can only change additional json header), this will not be there in the master file. It is more for the use of processor.

PR #173, opened 2020-09-08
server: Update mode to skip detector setup

thattil:
- update mode to skip detector setup for programming fpga (if user cant start server without errors)
- check kernel version for m3 and g2


slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -23,6 +23,7 @@
 
 // Global variable from slsDetectorServer_funcs
 extern int debugflag;
+extern int updateFlag;
```
erikfrojdh, line 26: I don't have a better name but could we name this so it tells that we are skipping the setup?

PR #179, opened 2020-09-10
g2 and m3: kernel checks only when its too old

thattil:


thattil:
> Would it be possible to not duplicate the date parsing function?

The date parsing is put into common.c. The rest of the checkKernelVersion is put into Nios (because the parsing from uname needs first word removed for Nios kernel version to get date only). Also used c api for uname as suggested.

Tested only with virtual (with code changes). Will test only tomorrow morning when I get access to a board.

slsDetectorServers/slsDetectorServer/src/nios.c
```diff
@@ -126,4 +129,50 @@ int mapCSP0(void) {
...
+int Nios_checkKernelVersion(char *expectedVersion) {
+    // extract kernel date string
+    struct utsname buf;
+    if (uname(&buf) == -1) {
+        LOG(logERROR, ("Could not get kernel version\n"));
+        return FAIL;
+    }
+
+    // remove first word (#version number)
+    const char *ptr = strchr(buf.version, ' ');
```
erikfrojdh, line 143: Are we sure that this search always works? On my machine it's the second whitespace + 1.

slsDetectorServers/slsDetectorServer/src/nios.c
```diff
@@ -126,4 +129,50 @@ int mapCSP0(void) {
...
+    if (ptr == NULL) {
+        LOG(logERROR, ("Could not parse kernel version\n"));
+        return FAIL;
+    }
+    char output[256];
+    memset(output, 0, 256);
```
erikfrojdh, line 149: Not that this is a performance critical part of the code, but could we pass just the pointer into the buf.version string instead of making a copy here

On the other hand since we access it multiple times why not.

slsDetectorServers/mythen3DetectorServer/slsDetectorFunctionList.c
```diff
@@ -167,19 +167,8 @@ int checkKernelVersion() {
...
-    pclose(sysFile);
 
-    if (strstr(output, KERNEL_DATE_VRSN) == NULL) {
-        LOG(logERROR, ("Kernel Version Incompatible! Expected: %s, Got: %s\n",
-                       KERNEL_DATE_VRSN, output));
-        return FAIL;
-    }
-    LOG(logINFO, ("Kernel Version Compatible: %s\n", output));
-    return OK;
+    return Nios_checkKernelVersion(KERNEL_DATE_VRSN);
```
erikfrojdh, line 171: Cool!

erikfrojdh (commented):
Looks good. Just check the parsing of the returned date.

slsDetectorServers/slsDetectorServer/src/nios.c
thattil: Yes, exactly. That's why i put this into Nios.c. Because in the actual nios board, it doesnt have the SMP,  just one word before.

PR #180, opened 2020-09-11
Adding auto generated server default #define in documentation

erikfrojdh:
For each detector parse slsDetectorServer_defs.h to extract the defines and write them to a CSV. 
Generate table from the CSV using sphinx. 
Now included in make docs


thattil (approved):
Looks good :)

PR #214, opened 2020-11-10
Defaultdacs

thattil:
jungfrau: removed default dacs from settings
all others except ctb: connected default dacs
gotthard2: connected on chip dacs as well


erikfrojdh:
Looks good, the only thing that itches is the repetition of the setDefaultDac function which seems to be the same for all except gotthard2.

PR #229, opened 2021-01-18
Pattern.h as a public header files

xiaoqiangwang:
It is used in Detector.h.


erikfrojdh:
Good catch, and sorry for missing the header. We are currently not building from the installed version in the CI pipeline, something that we need to start doing.

PR #252, opened 2021-06-10
added counters for m3 gui

thattil:
counters in m3 gui

@anberga 
Screenshot from 2021-06-11 09-23-48

### DIFF
--- a/slsDetectorGui/src/qTabDataOutput.cpp
+++ b/slsDetectorGui/src/qTabDataOutput.cpp
@@ -1,9 +1,7 @@
 #include "qTabDataOutput.h"
 #include "qDefs.h"
-#include "sls/bit_utils.h"
 
 #include <QButtonGroup>
-#include <QCheckBox>
 #include <QFileDialog>
 #include <QStandardItemModel>
 #include <QString>
@@ -25,12 +23,6 @@ void qTabDataOutput::SetupWidgetWindow() {
     btnGroupRate = new QButtonGroup(this);
     btnGroupRate->addButton(radioDefaultDeadtime, 0);
     btnGroupRate->addButton(radioCustomDeadtime, 1);
-    chkCounter1->setEnabled(false);
-    chkCounter2->setEnabled(false);
-    chkCounter3->setEnabled(false);
-    chkCounter1->hide();
-    chkCounter2->hide();
-    chkCounter3->hide();
 
     // enabling according to det type
     switch (det->getDetectorType().squash()) {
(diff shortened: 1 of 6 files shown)


slsDetectorGui/src/qTabDataOutput.cpp
```diff
@@ -416,6 +438,57 @@ void qTabDataOutput::SetParallel(bool enable) {
...
+    disconnect(chkCounter1, SIGNAL(toggled(bool)), this,
+               SLOT(SetCounterMask()));
+    disconnect(chkCounter2, SIGNAL(toggled(bool)), this,
+               SLOT(SetCounterMask()));
+    disconnect(chkCounter3, SIGNAL(toggled(bool)), this,
+               SLOT(SetCounterMask()));
+    try {
+        auto retval = sls::getSetBits(det->getCounterMask().tsquash(
+            "Counter mask is inconsistent for all detectors."));
+        std::vector<QCheckBox *> counters = {chkCounter1, chkCounter2,
```
erikfrojdh, line 452: I wonder if this could be done once when initializing? Keeping the vector as a member variable?

slsDetectorGui/src/qTabDataOutput.cpp
```diff
@@ -416,6 +438,57 @@ void qTabDataOutput::SetParallel(bool enable) {
...
+                                             chkCounter3};
+        // default to unchecked
+        for (unsigned int i = 0; i < counters.size(); ++i) {
```
erikfrojdh, line 455: Since you touch all elements maybe std::for_each is more clear?

slsDetectorGui/src/qTabDataOutput.cpp
```diff
@@ -416,6 +438,57 @@ void qTabDataOutput::SetParallel(bool enable) {
...
+            counters[i]->setChecked(false);
+        }
+        // if retva[i] = 2, chkCounter2 is checked
+        for (unsigned int i = 0; i < retval.size(); ++i) {
```
erikfrojdh, line 459: Maybe something like this?
```
for (auto i : retval){
   counters[i]->setChecked(True)
}
```

erikfrojdh (commented):
Looks ok, some small comments that might make it clearer...

slsDetectorGui/src/qTabDataOutput.cpp
thattil: done

slsDetectorGui/src/qTabDataOutput.cpp
thattil: done

slsDetectorGui/src/qTabDataOutput.cpp
thattil: used range based for loop as you suggested. thanks!

PR #256, opened 2021-06-30
Disabledata

thattil:
Disable data stream for Eiger

Rewritten file writing in receiver to separate matter and virtual

No file writing if not activated ( Todo)

No file writing if data stream disabled ( to do in receiver)


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -1369,6 +1369,15 @@ void Detector::setQuad(const bool enable) {
     pimpl->Parallel(&Module::setQuad, {}, enable);
 }
 
+Result<bool> Detector::getDataStream(const bool left, Positions pos) const {
+    return pimpl->Parallel(&Module::getDataStream, pos, left);
+}
+
+void Detector::setDataStream(const bool left, const bool enable,
```
erikfrojdh, line 1376: Enum for left, right, top, bottom? Then it would also work for other detectors.

PR #259, opened 2021-07-14
G2timejreset

thattil:
G2: bug fix for time parameters for tolerance . eg. 220ns should be rounded to 222ns, not 215ns
J: stopping trigger mode and switching back to auto mode, blocks data streaming. Temporary workaround in server to reset core in stop acquisition until firmware release.


erikfrojdh (approved):
LGTM (looks good to me)

PR #264, opened 2021-07-15
vetostream

thattil:
vetostream none, 3gbe, 10gbe, 3gbe 10gbe

10gbe will call setnumudpinterfaces that handles rxr udp interfaces and zmq interfaces
3gbe will call setvetostream in detector server (only3gbe)

Need to add python code


slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
```diff
@@ -2578,6 +2578,23 @@ int getVeto() {
             CONFIG_VETO_ENBL_OFST);
 }
 
+void setVetoStream(int value) {
```
erikfrojdh, line 2581: Does setVetoStream only consider the 3gbe interface?

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -1547,6 +1547,37 @@ void Detector::setVeto(bool enable, Positions pos) {
...
+    return res;
+}
+
+void Detector::setVetoStream(defs::EthernetInterface interface, Positions pos) {
+    // 3gbe
+    bool i3gbe = (interface & defs::EthernetInterface::I3GBE);
+    pimpl->Parallel(&Module::setVetoStream, pos, i3gbe);
+
+    // 10gbe (debugging interface) opens 2nd udp interface in receiver
+    int old_numinterfaces = getNumberofUDPInterfaces(pos).tsquash(
```
erikfrojdh, line 1573: Do we need this check or can we just set it to simplify the code?

erikfrojdh (commented):
Overall good, found it a bit confusing at first with the numinterfaces vs. setVetoStream but it's probably the way to do it.

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
thattil: In the detector server yes, because the setNumberofUDPInterfaces already does this for 10gbe and does a whole lot of other things related to it cuz of zmq etc.

slsDetectorSoftware/src/Detector.cpp
thattil: Setting numberofudpinterfaces each time seems a bit overkill if we do:
p vetostream 10gbe 3gbe
p vetostream 10gbe

because it does a whole bunch of things due to zmq

slsDetectorServers/gotthard2DetectorServer/slsDetectorFunctionList.c
erikfrojdh: I see

slsDetectorSoftware/src/Detector.cpp
erikfrojdh: ok

thattil:
> Overall good, found it a bit confusing at first with the numinterfaces vs. setVetoStream but it's probably the way to do it.

I agree, but we can change this in the future as it wlil be a big change. Along with that we can also see how we can make eiger into numinterfaces 2 as well (as alejandro suggested)

PR #267, opened 2021-07-22
Jungfrau features: 1. chip version

thattil:
- setting via config file in server (chip 1.1 is incompatible with HW 1.0)
- getting from client


slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
```diff
@@ -468,6 +477,119 @@ int setDefaultDacs() {
     return ret;
 }
 
+int readConfigFile() {
+
+    if (initError == FAIL) {
+        return initError;
+    }
+
+    char fname[128];
```
erikfrojdh, line 486: Do you want to use a constant for the size as below, to make sure there can be no mismatch?

slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
```diff
@@ -468,6 +477,119 @@ int setDefaultDacs() {
...
+    LOG(logINFOBLUE, ("Reading config file %s\n", CONFIG_FILE));
+
+    // Initialization
+    const size_t LZ = 256;
+    char line[LZ];
+    memset(line, 0, LZ);
+    char command[LZ];
+
+    // keep reading a line
+    while (fgets(line, LZ, fd)) {
```
erikfrojdh, line 511: Can some of the reading logic be merged into a common function for all detectors? I'm asking since many now have their own config files...

erikfrojdh (commented):
LGTM, one minor thing with the size which you can fix before merging

slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
thattil: Yes, but i was pushing it off for later due to the urgency.  Ill try to put it in for now if its fast. Else have to make a different PR for it.

slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
thattil: will do

slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
thattil: We decided to postpone this to another PR. Every detector has different commands and there are a bit too many to do it in this PR.

PR #268, opened 2021-07-23
Jungfrau features 2. config chip

thattil:
J: 2. configure chip [only for chip 1.1)
    • if chip is powered, configure chip (default configuration), else throw exception
    • if chip is powered off or at start up or fpgareset, chip is unconfigured
    • Before acquisition, check if chip is configured, else throw exception


erikfrojdh (commented):
LGTM!

PR #269, opened 2021-07-23
Jungfrau features: 3. settings

thattil:
J: 3. settings (for all chips)
    • Remove forceswitch1 and forceswitch2 from enums (Settings will be only dynamicgain, dynamichg0)
    • change 3 special dacs to defaults for that setting
    • [chip1.1] if powerchip on, configure chip 
    •changing a dac individually also validates setting (setting becomes undefined if the special dacs have different values)


slsDetectorGui/src/qTabMeasurement.cpp
```diff
@@ -213,23 +213,16 @@ void qTabMeasurement::SetupTimingMode() {
...
-            item[(int)GATED]->setEnabled(true);
-            item[(int)BURST_TRIGGER]->setEnabled(true);
-            break;
-        case slsDetectorDefs::MYTHEN3:
-            item[(int)GATED]->setEnabled(true);
-            item[(int)TRIGGER_GATED]->setEnabled(true);
-            break;
-        default:
-            break;
+        try {
```
erikfrojdh, line 218: Nice!

slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
```diff
@@ -884,19 +884,30 @@ enum detectorSettings setSettings(enum detectorSettings sett) {
     if (sett == UNINITIALIZED)
         return thisSettings;
 
+    const int specialDynamicDacValues[NSPECIALDACS] =
```
erikfrojdh, line 887: Could be moved into the switch statement to create the array we need.

erikfrojdh (commented):
Maybe move the array, otherwise go ahead and merge

slsDetectorGui/src/qTabMeasurement.cpp
thattil: Need to do the same for getdaclist etc, but this might not be so easy. added it to backlog

slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
thattil: Yes, i had incorrectly assumed declarations inside switch statements dont work

thattil: will fix it

PR #278, opened 2021-07-30
Jungfrau features: 8. storagecells

thattil:
[change only for chip1.1]
    • additional number of storage cells not applicable (only single acquisitions allowed (multiple frame/triggers allowed))
    • storage cell start is from 0 – 3 (each bit for each cell)
    • storagecell_delay not applicable


erikfrojdh (approved):
LGTM

PR #282, opened 2021-08-02
jungfrau features: 9. auto comp disable time

thattil:
• less % and not restrictive about long exptime, register to set how long comparator needs to be enabled.

- (ony for v1.1) new register to custom the period in ns. (chipv1.0,  it was 93 or so %)
- put it along with command line argument

### DIFF
--- a/slsDetectorSoftware/src/CmdProxy.cpp
+++ b/slsDetectorSoftware/src/CmdProxy.cpp
@@ -1729,6 +1729,37 @@ std::string CmdProxy::TemperatureEvent(int action) {
     return os.str();
 }
 
+std::string AutoComparatorDisable(const int action) {
+    std::ostringstream os;
+    os << cmd << ' ';
+    if (action == slsDetectorDefs::HELP_ACTION)
+        os << "[0, 1]\n\t[Jungfrau] Auto comparator disable mode. By default, "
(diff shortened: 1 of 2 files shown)


slsDetectorSoftware/include/sls/Detector.h
```diff
@@ -1131,15 +1131,23 @@ class Detector {
...
+     * automatically after 93.75% of exposure time (only for longer than 100us).
+     * The % is for chipv1.0. One can set the duration for chipv1.1 using
+     * setComparatorDisableTime\n Default is false or this mode
+     * disabled(comparator enabled throughout). true enables mode. 0 disables
+     * mode.
      */
     void setAutoCompDisable(bool value, Positions pos = {});
 
+    /** [Jungfrau] */
+    Result<ns> getComparatorDisableTime(Positions pos = {}) const;
```
erikfrojdh, line 1145: Should we change the name to getCompDisableTime? to be consistent with getAutoCompDisable?

erikfrojdh (commented):
Looks good.

slsDetectorSoftware/include/sls/Detector.h
thattil: done. the other way. change getAutoCompDisable to getAutoComparatorDisable.

PR #283, opened 2021-08-03
Jungfrau features: 11. update

thattil:
- settings: remove check for special dacs, should not return undefined
 [g0, hg0]

- gain mode:
[dynamicgain, forceswitchg1, forceswitchg2, fixg1, fixg2, fixg0]
fixg0 and fixhg0 is combined to fixg0 as it depends on settings (g0 and hg0)

- fixg0 is cmp_reset bit and requires a warning in help that they can damage the detector when using this without caution
- also disabled unless in advanced mode in gui, warning when setting it

- any command touching the chip1.1 register will not check the status when setting. It will just set and assume that it has been set.

### DIFF
--- a/python/src/enums.cpp
+++ b/python/src/enums.cpp
@@ -308,7 +308,7 @@ void init_enums(py::module &m) {
         .export_values();
 
     py::enum_<slsDetectorDefs::gainMode>(Defs, "gainMode")
-        .value("DYNAMIC_GAIN", slsDetectorDefs::gainMode::DYNAMIC_GAIN)
+        .value("DYNAMIC_GAIN_MODE", slsDetectorDefs::gainMode::DYNAMIC_GAIN_MODE)
         .value("FORCE_SWITCH_G1", slsDetectorDefs::gainMode::FORCE_SWITCH_G1)
         .value("FORCE_SWITCH_G2", slsDetectorDefs::gainMode::FORCE_SWITCH_G2)
         .value("FIX_G1", slsDetectorDefs::gainMode::FIX_G1)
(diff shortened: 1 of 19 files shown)


python/src/enums.cpp
```diff
@@ -307,8 +308,11 @@ void init_enums(py::module &m) {
         .export_values();
 
     py::enum_<slsDetectorDefs::gainMode>(Defs, "gainMode")
-        .value("NORMAL_GAIN_MODE", slsDetectorDefs::gainMode::NORMAL_GAIN_MODE)
+        .value("DYNAMIC_GAIN_MODE", slsDetectorDefs::gainMode::DYNAMIC_GAIN_MODE)
```
erikfrojdh, line 311: Do we now change this to DYNAMIC_GAIN?

slsDetectorGui/include/qTabSettings.h
```diff
@@ -52,9 +58,21 @@ class qTabSettings : public QWidget, private Ui::TabSettingsObject {
         G2_LOWCAP_LOWGAIN,
         G4_HIGHGAIN,
         G4_LOWGAIN,
+        GAIN0,
         UNDEFINED,
         UNINITIALIZED,
         NUMSETTINGS
     };
+
+    enum {
```
erikfrojdh, line 67: Is it possible to reuse the enums from sls_detector_defs.h instead of defining new ones?

thattil: I tried this, but then undefined and uninitialized has enum values 200.. and i wanted this to get the values from the drop down

PR #284, opened 2021-08-05
Jungfrau features: 13. flipped data x

thattil:
- flippeddatax [0|1] - 1 flips bottom across x axis only for HW2.0, only sets detector server, (slsReceiver and gui has no effect)

- changed API from setBottom to setFlippedDataOverXAxis


slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
```diff
@@ -2002,6 +2003,25 @@ void alignDeserializer() {
           bus_r(ADC_DSRLZR_3_REG) & (~(ADC_DSRLZR_3_RFRSH_ALGNMNT_MSK)));
 }
 
+int getFlipRows() {
+    return ((bus_r(CONFIG_REG) & CONFIG_BOTTOM_INVERT_STREAM_MSK) >>
+            CONFIG_BOTTOM_INVERT_STREAM_OFST);
+}
+
+void setFlipRows(int arg) {
+    if (arg >= 0) {
```
erikfrojdh, line 2012: What happens if the arg is smaller than 0?

Could actually rewrite the current as 

```
if (arg){
}else{
}

erikfrojdh (commented):
Go ahead!

slsDetectorServers/jungfrauDetectorServer/slsDetectorFunctionList.c
thattil: thats the point. If it < 0, it will not set flip rows.. so in the client interface, when it validates, it will fail.

PR #293, opened 2021-08-18
Roundrobin

thattil:
- udp_numdst. udp_dstlist,udp_firstdst (done)
- eiger (refactored server, also ip/mac was converted to string and back to int earlier, now fixed)
- g2 and m3 (to be done for later, not connected in firmware)
- default values for ip, mac and port (port defaults, others from entry 0)
- virtual server updated for rr


slsSupportLib/include/sls/sls_detector_defs.h
```diff
@@ -457,35 +461,58 @@ typedef struct {
...
 
         bool operator==(const currentSrcParameters &other) const {
-            return ((enable == other.enable) && (fix == other.fix) &&
-                    (normal == other.normal) && (select == other.select));
+            return ((enable_ == other.enable_) && (fix_ == other.fix_) &&
+                    (normal_ == other.normal_) && (select_ == other.select_));
+        }
+    } __attribute__((packed));
+
+    struct udpDestination {
```
erikfrojdh, line 497: I think we can make it's own file for the udpDestination. If you're not using it in the server it doesn't even need to be pure c. Then we can go for using the ip and mac classes.

erikfrojdh (commented):
A bit big for me to look into detail but I think we can use the classes in the udpDest stuct.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1516,12 +1517,12 @@ void calcChecksum(udp_header *udp) {
...
-    int dstport = udpDetails.dstport;
+    uint32_t srcip = udpDetails[0].srcip;
+    uint32_t dstip = udpDetails[0].dstip;
+    uint64_t srcmac = udpDetails[0].srcmac;
+    uint64_t dstmac = udpDetails[0].dstmac;
+    int srcport = udpDetails[0].srcport;
+    int dstport = udpDetails[0].dstport;
 
     LOG(logINFOBLUE, ("Configuring MAC\n"));
     char src_mac[50], src_ip[INET_ADDRSTRLEN], dst_mac[50],
```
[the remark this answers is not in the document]
thattil: Changed to 18 bytes

slsSupportLib/include/sls/sls_detector_defs.h
thattil: done

PR #295, opened 2021-09-01
Modid

thattil:
-serialnumber detached from eiger
-moduleid get for eiger, m3 and g2

they all use detid_[detectortype] files. (eiger uses detid.txt for real hardware as it was in production).
get will get from file. ( set into register at server startup)


slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
```diff
@@ -322,15 +302,22 @@ u_int32_t getDetectorIP() {
 void initControlServer() {
     LOG(logINFOBLUE, ("Configuring Control server\n"));
     if (!updateFlag && initError == OK) {
-        readDetectorNumber();
+#ifndef VIRTUAL
```
erikfrojdh, line 305: Why is there a difference between virtual and normal for getting the modid?

slsDetectorServers/eigerDetectorServer/slsDetectorServer_defs.h
```diff
@@ -2,7 +2,11 @@
 #include "sls/sls_detector_defs.h"
 
 #define REQUIRED_FIRMWARE_VERSION (27)
-#define IDFILECOMMAND             "more /home/root/executables/detid.txt"
+#ifdef VIRTUAL
```
erikfrojdh, line 5: Add a comment, why the filenames differ

erikfrojdh (commented):
Looks good

slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
thattil: We can set module id only in Beb, which is not there in the virtual server. So, I just getModuleId as well but do not assign it to anything, as there is nothing to set for virtual servers.

thattil: Or should i set it to a variable which is picked up when streaming out udp packets?

slsDetectorServers/eigerDetectorServer/slsDetectorServer_defs.h
thattil: added the comment

PR #296, opened 2021-09-09
Progfpga

thattil:
@aldo 128k->4k?
check pointer moves in flash
compare pofs (using hexdump into file and kdiff3)


slsDetectorServers/ctbDetectorServer/CMakeLists.txt
```diff
@@ -18,12 +18,14 @@ add_executable(ctbDetectorServer_virtual
...
     ../slsDetectorServer/src/sharedMemory.c
     ../slsDetectorServer/src/loadPattern.c
+    ../../slsSupportLib/opensslMd5/md5.c
 )
 
 include_directories(
     ../slsDetectorServer/include
     ../../slsSupportLib/include
     ../../slsDetectorSoftware/include/sls/
+    ../../slsSupportLib/opensslMd5
```
erikfrojdh, line 28: maybe we can put the header in include/sls/md5/md5.h and the source in src/md5/md5.c? Then we wouldn't need the additional include path.

slsDetectorSoftware/CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ set(SOURCES
     src/CmdProxy.cpp
     src/CmdParser.cpp
     src/Pattern.cpp
+    ../slsSupportLib/opensslMd5/md5.c
```
erikfrojdh, line 8: I think we should compile md5 as a part of the supportLib and then link it, instead of adding the sources here.

slsDetectorSoftware/CMakeLists.txt
```diff
@@ -13,6 +14,7 @@ add_library(slsDetectorObject OBJECT
 
 target_include_directories(slsDetectorObject PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../slsSupportLib/opensslMd5>"
```
erikfrojdh, line 17: Having already built the md5 as a part of the support lib we could remove the include here.

slsDetectorSoftware/src/Module.cpp
```diff
@@ -3404,57 +3406,54 @@ sls_detector_module Module::readSettingsFile(const std::string &fname,
     return myMod;
 }
 
+std::string Module::calculateChecksum(char *buffer, ssize_t bytes) {
```
erikfrojdh, line 3409: I think we can put this in it's own class, then it's easier to test since it' doesn't depend on the Module. Same folder as md5.c?

slsSupportLib/opensslMd5/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
erikfrojdh, line 1: Should we mention that this applies only for the md5.h/md5.c?

erikfrojdh (commented):
I would make some changes to the structure of the files and CMake build otherwise looks good. Didn't look too carefully on the server side.

PR #298, opened 2021-09-15
Somefixes

thattil:
- connected module id for m3
- check if detector idle before flipping rows in jungfrau detector
- rename loadsettings to loadtrimbits, also setmodule only used for only for m3 and eiger


slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -1503,7 +1503,7 @@ int set_module(int file_des) {
     ret = OK;
     memset(mess, 0, sizeof(mess));
 
-#if defined(CHIPTESTBOARDD) || defined(MOENCHD) || defined(GOTTHARD2D)
+#if !(defined(MYTHEN3D) && defined(EIGERD))
```
erikfrojdh, line 1506: Should probably be an || ? As it stands now both Mythen3 and Eiger has to be defined...

erikfrojdh (changes requested):
One fix then merge.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
thattil: Thanks!

PR #304, opened 2021-10-06
changing ports only in shared memory and not going to detector/receiv…

thattil:
changing ports only in shared memory and not going to detector/receiver server to change current tcp port


erikfrojdh (commented):
Also good, you can merge!

PR #321, opened 2021-10-19
slsDetectorServer: fix checks on UDP destination entry range

alejandrohomsp:
Hi,

What do you think about the checks around `MAX_UDP_DESTINATION` in this patch?

Cheers

Alejandro


thattil:
@alejandrohomsp The check has to be from 0 to 31. Entry 0 will never reach this function in the server as the client itself calls the traditional functions to set src ip, src mac, dst ip etc. The check is just to ensure this. So, in effect set_dest_udp_list is only allowed for 1-31, whereas get_dest_udp_list, I allow for 0-31.

Ofcourse, this will change once we have feedback from the other branch rr_rxr.

thattil:
Hi Alejandro,

That is a very good catch! Thanks a lot. Will merge this

Cheers,
Dhanya

alejandrohomsp:
Thank you @thattil for your work in `Round-Robin` implementation!

PR #330, opened 2021-10-29
Program firmware for new kernel

thattil:
using gpio3 pins to enforce proper programming.
checking kernel version before using the pins


erikfrojdh (commented):
Looks good but as always trust me less when it's about the server. 

One thing we could think about is to break out common functions, like removing a substring, to it's own library... but later.

thattil:
> Looks good but as always trust me less when it's about the server.
> 
> One thing we could think about is to break out common functions, like removing a substring, to it's own library... but later.

Totally agree! There are other string manipulations like safely copying a string into another. Something like your string_utils.cpp. But yes, adding onto the list of things to do.

PR #333, opened 2021-11-10
3kernelupdate

thattil:
- updatekernel
-updatedetectorserver [copydetectorserver deprecated]
-update (without tftp) added  [with tftp deprecated]
- programfpga (refactored to reuse code for updatekernel/updatedetectorserver/programfpga)
- execute commands print module id and hostname when it fails as well
- change system commands to executecommand in server side
- more error checking in basic tests
- file_utils moved to sls namespace


CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ sls_enable_cxx_warning("-Wduplicated-cond")
 sls_disable_cxx_warning("-Wclass-memaccess")
 sls_disable_c_warning("-Wstringop-truncation")
 sls_disable_c_warning("-Wformat-nonliteral")
+sls_disable_c_warning("-Wformat-truncation")
```
erikfrojdh, line 141: In principle we don't need to disable this warning. It only warns if the formatting can be truncated and this is not acted on. 

Code like this doesn't generate a warning:
```
int r = snprintf(mess, MAX_STR_LENGTH-100, "Could not get kernel version. %s\n",
                 retvals);
if (r>MAX_STR_LENGTH)
     mess[MAX_STR_LENGTH-2] = '.';
```

In principle we could have a function that adds "..." when the string is truncated. Will it happen? Is it important? Nor sure =)

CMakeLists.txt
```diff
@@ -138,6 +138,7 @@ sls_enable_cxx_warning("-Wduplicated-cond")
 sls_disable_cxx_warning("-Wclass-memaccess")
 sls_disable_c_warning("-Wstringop-truncation")
 sls_disable_c_warning("-Wformat-nonliteral")
```
erikfrojdh, line 140: It is also quite easy to enable this warning. I didn't find anywhere where we reuse the char* to format. It's not 100% related to this PR but having this warning on might be nice.

slsDetectorServers/slsDetectorServer/src/common.c
```diff
@@ -370,12 +402,177 @@ int verifyChecksum(char *mess, char *clientChecksum, MD5_CTX *c, char *msg) {
...
+}
+
+int moveBinaryFile(char *mess, char *dest, char *src, char *errorPrefix) {
+    char cmd[MAX_STR_LENGTH] = {0};
+    char retvals[MAX_STR_LENGTH] = {0};
+
+    // one can move into the current process binary (will not interfere in
+    // kernel mode)
+
+    char *format = "mv %s %s";
```
erikfrojdh, line 563: This could be put directly in the snprintf since it's not reused. Advantage is that the warning can be turned on

erikfrojdh (commented):
Some minor things about the two warnings. The auto formatting gave quite large diffs so not sure I caught everything else///

CMakeLists.txt
thattil: I added this before 6.0.0 was released. It was to do with the snprintf combining mess and retvals that were both the same size, even though we were dealing with it

slsDetectorServers/slsDetectorServer/src/common.c
thattil: will check it out

thattil: ok done

CMakeLists.txt
thattil: removed both warnings and fixed it in places.

PR #346, opened 2022-01-24
CMake integration for Moench

erikfrojdh:
Removed the makefiles and integrated the Moench binaries in the normal CMake build. Added short option to build only Moench executables. 

```
cmake ../slsDetectorPackage -DSLS_BUILD_ONLY_MOENCH=ON
```


erikfrojdh:
Changes already merged by other PR

PR #349, opened 2022-01-28
Moenchfnum

thattil:
- this PR is for both ctb and moench
- set frame number for next acquisition 
- adjust in virtual servers for the fake frame numbers that can be set
- adjust in 1g udp interface to allow to set frame numbers
- set default starting number at server start up


RELEASE.txt
```diff
@@ -26,7 +26,11 @@ This document describes the differences between v6.1.0 and v6.0.0.
 - refactoring (rxr)
 - fixed patsetbit and patsetmask for moench
 - changed default vref of adc9257 to 2V for moench (from 1.33V)
+<<<<<<< HEAD
```
erikfrojdh, line 29: I guess this should not be in the commit? Looks like the traces of a merge conflict?

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -898,6 +898,24 @@ int getReadoutMode() {
...
+    setU64BitReg(value, NEXT_FRAME_NUMB_LOCAL_LSB_REG,
+                 NEXT_FRAME_NUMB_LOCAL_MSB_REG);
+#ifndef VIRTUAL
+    // for 1g udp interface
+    setUDPFrameNumber(value);
+#endif
+    return OK;
+}
+
+int getNextFrameNumber(uint64_t *retval) {
```
erikfrojdh, line 913: What will the virtual server return?

thattil: what ever is in this register.. it is set up to be 1 at startup..

RELEASE.txt
thattil: Yes it does. thanks for seeing that. ill fix it soon

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: the return is a ok or fail (useful only for eiger).. but the argument will return whatever is in the register for the next frame number.

PR #355, opened 2022-01-28
Rxframeindex

thattil:
- increases the progress according to listened index. (not processed index)
- current frame index points to listened frame index (not processed index)
- when in discard partial frames or empty mode, the frame number doesnt increase by 1, it increases to that number (so its faster)


erikfrojdh (approved):
LGTM

PR #362, opened 2022-02-02
Eiger datastream

thattil:
- eiger server: fix for datastream enabling wrong ports for bottom half module
- receiver suimmary print out more clear when it is deactivated
- exception when trying to disable data stream in 10g mode


erikfrojdh (approved):
LGTM!

PR #371, opened 2022-02-04
Gotthard25um: gui image

thattil:
- image reconstruction for gottard25um in gui
- first module (master) interleaves with second module (slave), the first channel being masters first and the second channel being the slaves last channel.
- only 2 modules allowed max for a gotthard1 or 2 detector type


erikfrojdh (approved):
Looks good. Long term we should maybe have the data manipulation function somewhere outside the GUI

thattil:
> Looks good. Long term we should maybe have the data manipulation function somewhere outside the GUI

I agree :+1:

PR #376, opened 2022-02-07
eiger datastream fix for 1g

thattil:
datastream is only for 10g for eiger atm, its mentioned now  in comments. accordingly handled in receiver. 

This is better solution, in case it was disabled in 10g, and not possible to set enable in 1g mode, which is given to the receiver. (in fixing this bug)


erikfrojdh (approved):
LGTM

PR #378, opened 2022-02-09
localhostMac

thattil:
clearer exceptions to help user fix the issue of localhost not getting mac address


thattil (commented):
Erik: Check Detector Virtual server at "setudpip" if localhost is rx_hostname and udp_destip is auto. Remove this check. Change ethname to macadress function to include "lo" giving "00:00:00:00:00:00"

erikfrojdh (approved):

PR #379, opened 2022-02-09
M3gaincaps

thattil:
-reg for m3 is reserved only for gaincaps and not settings. Fixed in set threshold and setall threshold
@anberga


erikfrojdh (approved):
LGTM

PR #395, opened 2022-02-22
setmaster

thattil:
- Allowing set master for eiger
- getmaster can give error for eiger


thattil:
Replaced with PR #397

PR #396, opened 2022-02-22
updatemode

thattil:
hostname command failed when connecting to servers in update mode


slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -99,7 +98,12 @@ int updateModeAllowedFunction(int file_des) {
                             F_UPDATE_KERNEL,
                             F_UPDATE_DETECTOR_SERVER,
                             F_GET_UPDATE_MODE,
-                            F_SET_UPDATE_MODE};
+                            F_SET_UPDATE_MODE,
+                            F_GET_NUM_CHANNELS,
+                            F_GET_NUM_INTERFACES,
+                            F_ACTIVATE};
+    unsigned int listsize = 22;
```
erikfrojdh, line 105: Sorry, this is in principle not related to the PR since the line didn't change. But, I couldn't help myself because it looks like a bug waiting to happen =) 

```
unsigned int listsize = 22;
```
can be replaced with
```
size_t listsize = sizeof(list)/sizeof(enum detFuncs);
```
Maybe we also want to change the name of the list variable since it technically isn't a list =) allowed_funcs?

erikfrojdh (changes requested):
Looks good but I would still do the fix.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
thattil: makes sense

thattil: Thanks! Done! :)

PR #397, opened 2022-02-23
Setmaster

thattil:
- connect master commands to api (allow set master for eiger)
- get master in server can give errors
- --ignore-config command line
- command line argument 'master' mainly for virtual servers (also master/top for real eiger)
- only one virtual server for eiger, use command lines for master/top
- all stop servers also check for errors at startup( in case it was running with an older version)


erikfrojdh (approved):
Go for it

PR #398, opened 2022-02-24
Receiver frame indices and progress

thattil:
- framescaught and frameindex now returns a vector for each port
- progress looks at activated or enabled ports, so progress does not stagnate
- (eiger) disable datastreaming also for virtual servers only for 10g
- missing packets also takes care of disabled ports


erikfrojdh (approved):
Looks good. Go ahead and merge.

PR #407, opened 2022-03-15
Stopfnum

thattil:
eiger: get next frame number for 10g was connected to 1g registers (always gave incorrect values)
eiger/jungfrau/ctb/moench: for any stop acquisition, check if next frame number is inconsistent, if so, set it to the max + 1 value


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -800,6 +800,25 @@ void Detector::startDetectorReadout() {
 
 void Detector::stopDetector(Positions pos) {
     pimpl->Parallel(&Module::stopAcquisition, pos);
+
+    // validate consistent frame numbers
+    switch (getDetectorType().squash()) {
```
erikfrojdh, line 805: Do we need the switch? The others don't support setting frame number?

erikfrojdh (approved):
one question but go ahead and merge

slsDetectorSoftware/src/Detector.cpp
thattil: Yes. Only Eiger, Jungfrau, ctb and moench have it implemented in the firmware. Ctb and moench were recent additions

PR #414, opened 2022-03-24
Specialfile

thattil:
- nios: check if drive is special file before continuing to flash, if it is normal file, suggest reboot
- blackfin: 
   - kernel: check if drive is special file before continuing to flash, if it is normal file, suggest reboot
   - fpga: check if drive is special, if it is normal, suggest --force-delete-normal-file
   - if the argument is there, delete normal file, create drive file, continue programming


slsDetectorServers/slsDetectorServer/src/programViaBlackfin.c
```diff
@@ -38,6 +39,9 @@
 
 #define CMD_GET_AMD_FLASH  "dmesg | grep Amd"
 
+#define CMD_CREATE_DEVICE_FILE_PART1 "mknod"
+#define CMD_CREATE_DEVICE_FILE_PART2 "c 90 6"
+
 #define FLASH_BUFFER_MEMORY_SIZE (128 * 1024) // 500 KB
```
erikfrojdh, line 45: This is why I'm a bit sceptical to comments =)

erikfrojdh (approved):
Looks good, let's bring it in to developer.

PR #421, opened 2022-03-31
M3: polarity, interpolation, pump probe

thattil:


python/slsdet/detector.py
```diff
@@ -3498,10 +3498,57 @@ def clkfreq(self):
...
     def readout(self):
         """
-        Mythen3] Starts detector readout. Status changes to TRANSMITTING and automatically returns to idle at the end of readout.
+        [Mythen3] Starts detector readout. Status changes to TRANSMITTING and automatically returns to idle at the end of readout.
         """
         self.startDetectorReadout()
     
+    @property
+    @element
+    def polarity(self):
```
erikfrojdh, line 3507: Polarity enum is missing from python. You can get it by running python generate_enums.py from the same folder as the function script.

slsDetectorServers/mythen3DetectorServer/mythen3.c
```diff
@@ -149,54 +113,125 @@ patternParameters *setChipStatusRegisterPattern(int csr) {
...
+        csr |= 1 << _CSR_C15pre;
+
+    return setChipStatusRegister(csr);
+}
+
+int getInterpolation() {
+    return ((chipStatusRegister & CSR_interp_MSK) >> CSR_interp);
+}
+
+int setInterpolation(int mask) {
```
erikfrojdh, line 161: Can we come up with a better name than mask?

erikfrojdh (changes requested):
Few comment. Will have a second look when the I can test with python.

python/slsdet/detector.py
thattil: done

python/slsdet/detector.py
```diff
@@ -3498,10 +3498,57 @@ def clkfreq(self):
...
+        """[Mythen3] Set positive or negative polarity. Enum: polarity"""
+        return self.getPolarity()
+
+    @polarity.setter
+    def polarity(self, value):
+        ut.set_using_dict(self.setPolarity, value)
+
+    @element
```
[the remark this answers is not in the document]
thattil: done

slsDetectorServers/mythen3DetectorServer/mythen3.c
thattil: done

PR #423, opened 2022-04-04
copy detector server: rm old server 

thattil:
- removing old server or target of linked file when copying a detector server (via both tftp and not) for blackfin detectors (less space)
- clearing up abs path and respawn path (double // etc)
- refactoring booting up macros for readability
- if servername copied is same as final link name, raise error
- hostname error causes generic detector type in shm in multi, programfpga then shoudl have clearer error message.
- removed copydetectorserver


erikfrojdh (approved):
LGTM!

PR #425, opened 2022-04-04
sls_receiver_header* in callbacks

thattil:
char* to sls_receiver_header* in receiver data call backs


slsReceiverSoftware/include/sls/Receiver.h
```diff
@@ -51,8 +51,7 @@ class Receiver : private virtual slsDetectorDefs {
      * users get data to write depending on call backs registered
      */
     void registerCallBackStartAcquisition(int (*func)(std::string, std::string,
```
erikfrojdh, line 53: Would it work with a const std::string&? This could be more efficient depending on what the users want to do with the names.

slsReceiverSoftware/src/DataProcessor.cpp
```diff
@@ -316,16 +316,17 @@ uint64_t DataProcessor::ProcessAnImage(char *buf) {
...
                                      sizeof(sls_receiver_header),
-                                 (uint32_t)(*((uint32_t *)buf)), pRawDataReady);
+                                 dsize, pRawDataReady);
         }
 
         // call back with modified size
         else if (rawDataModifyReadyCallBack != nullptr) {
-            auto revsize = (uint32_t)(*((uint32_t *)buf));
-            rawDataModifyReadyCallBack((char *)rheader,
+            size_t revsize = (uint32_t)(*((uint32_t *)buf));
```
erikfrojdh, line 328: Would this be more clear? 
```
std::size_t revsize = *reinterpret_cast<uint32_t*>(buf);
```

erikfrojdh (commented):
I think we should try to get rid of the old style casts. Many of them are probably also undefined behavior, but very likely to work. Can be replaced with something like std::bitcast (C++20) but maybe we have other options.

slsReceiverSoftware/include/sls/Receiver.h
thattil: It would work, but I am not aware if the users are making copies to manipulate the strings.

thattil: I can change to this though, as the norm would be to not manipulate the strings I assume. The user gets to choose fpath, fname and findex.

slsReceiverSoftware/src/DataProcessor.cpp
thattil: done for both the casts.. and static for the startacquisitioncallback imagesize

slsReceiverSoftware/include/sls/Receiver.h
thattil: done

PR #436, opened 2022-04-26
Fixpositions

thattil:
- fixed row column indexing (mainly for multi module Jungfrau 2 interfaces )
- eiger gui row indices not flipped anymore (fix in config)
- using Y and X for row and column for detPos in server side


slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
```diff
@@ -2548,8 +2548,18 @@ void *start_timer(void *arg) {
                 }
                 break;
             case 16:
+#ifdef TEST_MOD_GEOMETRY
```
erikfrojdh, line 2551: Should this be in the final code? if so maybe a comment what the ifdef does?

erikfrojdh (approved):
Looks god.

slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
thattil: added comment

PR #439, opened 2022-04-28
Support external build of python lib

erikfrojdh:
Support building only the python libraries by calling

```
cmake path/to/slsDetectorPackage/python/
```

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,6 +1,17 @@
 # SPDX-License-Identifier: LGPL-3.0-or-other
 # Copyright (C) 2021 Contributors to the SLS Detector Package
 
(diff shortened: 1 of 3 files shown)


python/CMakeLists.txt
```diff
@@ -1,6 +1,17 @@
 # SPDX-License-Identifier: LGPL-3.0-or-other
 # Copyright (C) 2021 Contributors to the SLS Detector Package
 
+if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    #if we are only building the python extension
+    cmake_minimum_required(VERSION 3.12)    
+    project(slsdet)
+    find_package(slsDetectorPackage 6 REQUIRED)
```
thattil, line 8: is it find package 6.1.1 instead of 6?

thattil (approved):

PR #496, opened 2022-07-13
Rx: refactor memory structure and listener

thattil:
- refactoring listening udp socket code: deactivated and datastream dont create udp sockets anyway
- minor refactoring for readability (size_t for fifo size)
- gui message doesnt show if '>' symbol in error message
- refactoring of memory structure access in Thread objects
- empty images did not have version and det type in file write
- image structure instead of using offset increments
- listener using "port disabled" instead of if (!activate) || (!detectorDataStream) || (noRoi)
- listener: usleep after setting udpSocketAlive to false, so there's time before udpsocket shutdown (tcp thread at stopReceiver) and listening to udp socket (in listener thread)


slsReceiverSoftware/src/Listener.cpp
```diff
@@ -347,284 +319,174 @@ uint32_t Listener::ListenToAnImage(char *buf) {
...
+        break;
+    }
+    dstHeader.detHeader.packetNumber = numpackets; 
+    // for empty frames (padded)
+    if (isHeaderEmpty) {
+        dstHeader.detHeader.frameNumber = currentFrameIndex;
+        // no packet to get bnum
+        dstHeader.detHeader.row = row;
+        dstHeader.detHeader.column = column;
+        dstHeader.detHeader.detType = (uint8_t)generalData->myDetectorType;
```
erikfrojdh, line 419: Do you want to use static_cast?

erikfrojdh (commented):
Might have missed something due to the large diffs. But looks better to me. 
Let's see how testing goes.

slsReceiverSoftware/src/Listener.cpp
thattil: done

slsReceiverSoftware/src/BinaryDataFile.cpp
```diff
@@ -77,37 +75,34 @@ void BinaryDataFile::WriteToFile(char *buffer, const int buffersize,
...
     if (sizeof(sls_bitset) == sizeof(bitset_storage)) {
-        ret = fwrite(buffer, 1, buffersize, fd_);
+        ret = fwrite((char*)&header, 1, sizeof(sls_receiver_header) + imageSize, fd_);
     }
 
     // not contiguous bitset
     else {
         // write detector header
-        ret = fwrite(buffer, 1, sizeof(sls_detector_header), fd_);
+        ret = fwrite((char*)&header, 1, sizeof(sls_detector_header), fd_);
```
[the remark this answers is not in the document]
thattil: Yes, you are right. Its not needed when there isnt any offset. Ive removed the char* cast and also exchanged the size and count.

PR #498, opened 2022-07-13
Make #define from sls_detector_defs.h available in Python

erikfrojdh:
Added a script that parses sls_detector_defs.h and extracts the #defines and puts them in defines.py

**Usage**
```python
>>> import slsdet
>>> slsdet.LOCALHOST_IP
'127.0.0.1'
```


thattil (approved):
Niice!

PR #501, opened 2022-07-18
random data from virtual server in M3

erikfrojdh:
1. fix for compiling with gcc 12 (added 2x include <array>)
2. Generate random data with virtual mythen3 server to see the update in the gui


thattil (approved):
LGTM

thattil:
@erikfrojdh Do you still need this? Or can we close the PR?

thattil (changes requested):
Will need to be udpated if it needs to be merged.

PR #503, opened 2022-07-21
Rxclassmembers

thattil:
Rx
- same name arguments in constuctors
- class members suffix '_' removed 
- removed file.cpp


slsReceiverSoftware/src/BinaryDataFile.cpp
```diff
@@ -4,88 +4,92 @@
...
+        if (nullptr == (fd = fopen((const char *)fileName.c_str(), "wx"))) {
+            fd = nullptr;
             throw RuntimeError("Could not create/overwrite file " +
-                                    fileName_);
+                                    fileName);
         }
-    } else if (nullptr == (fd_ = fopen((const char *)fileName_.c_str(), "w"))) {
-        fd_ = nullptr;
-        throw RuntimeError("Could not create file " + fileName_);
+    } else if (nullptr == (fd = fopen((const char *)fileName.c_str(), "w"))) {
```
erikfrojdh, line 59: same as above

slsReceiverSoftware/src/BinaryDataFile.cpp
```diff
@@ -4,88 +4,92 @@
...
+    const std::string fPath, const std::string fNamePrefix,
+    const uint64_t fIndex, const bool ovEnable, const bool sMode,
+    const int modulePos, const int nUnitsPerReadout,
+    const uint32_t uPortNumber, const uint32_t mFramesPerFile) {
+
+    subFileIndex = 0;
+    numFramesInFile = 0;
+
+    filePath = fPath;
+    fileNamePrefix = fNamePrefix;
```
erikfrojdh, line 32: Do we need both fileNamePrefix and fNamePrefix? It looks like fNamePrefix is just used to set fileNamePrefix...

Same question applies to the other members.

slsReceiverSoftware/src/DataProcessor.cpp
```diff
@@ -28,20 +28,8 @@ namespace sls {
...
-                             std::vector<int> *ctblist, int *ctboff,
-                             int *ctbad)
-    : ThreadObject(index, typeName), fifo(f), detType(dType),
-      dataStreamEnable(dse),
-      streamingFrequency(sf),
-      streamingTimerInMs(st),
-      streamingStartFnum(sfnum), framePadding(fp),
-      ctbDbitList(ctblist), ctbDbitOffset(ctboff),
-      ctbAnalogDataBytes(ctbad) {
+DataProcessor::DataProcessor(int index, detectorType detType, Fifo *fifo, bool *dataStreamEnable, uint32_t *streamingFrequency, uint32_t *streamingTimerInMs, uint32_t *streamingStartFnum, bool *framePadding, std::vector<int> *ctbDbitList, int *ctbDbitOffset, int *ctbAnalogDataBytes)
```
erikfrojdh, line 31: Did you run clang format on the document?

erikfrojdh (commented):
Some small changes. For the renaming I didn't check every single line, but the compiler probably did =)

slsReceiverSoftware/src/BinaryDataFile.cpp
```diff
@@ -4,88 +4,92 @@
...
-    if (!overWriteEnable_) {
-        if (nullptr == (fd_ = fopen((const char *)fileName_.c_str(), "wx"))) {
-            fd_ = nullptr;
+    os << filePath << "/" << fileNamePrefix << "_d"
+       << (detIndex * numUnitsPerReadout + index) << "_f" << subFileIndex
+       << '_' << fileIndex << ".raw";
+    fileName = os.str();
+
+    if (!overWriteEnable) {
+        if (nullptr == (fd = fopen((const char *)fileName.c_str(), "wx"))) {
```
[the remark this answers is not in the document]
thattil: Thanks! Done.

slsReceiverSoftware/src/BinaryDataFile.cpp
thattil: Done

slsReceiverSoftware/src/DataProcessor.cpp
thattil: Nope, because there were too many things out of format. and it will be a nightmare for you to see the difference. I thought of running this directly on developer branch after this and the next PR and ofcourse, then to make sure this is done for every PR.

slsReceiverSoftware/src/BinaryDataFile.cpp
thattil: one is a data member and the other is an actual argument used to set it upon the first file creation. For consecutive file creations in the same acquisition, the data member is then used.

erikfrojdh: My bad looked quickly and mistook it for the constructor. I wonder if we could pass these arguments to the constructor and then do the creation by calling CreateFille when needed. But that might mess up the other structure...

slsReceiverSoftware/src/DataProcessor.cpp
erikfrojdh: Sounds good!

slsReceiverSoftware/src/BinaryDataFile.cpp
thattil: Well, I am hoping in the next PR to use the set methods to set up these members when the class objects are created and when these parameters are changed from the client. So we would be throwing away quite a few parameters. Also, planning to set one value instead of  fname, detindex and uniterperreadout as this string will not change for a particular thread.

PR #510, opened 2022-08-09
G2hdi

thattil:
-change HDI for new HDI version for 25 um:
HDI version 1.0:
master ID : 683 (R=20 kOhm)
slave ID: 704 (R=22 kOhm)
HDI version 2.0:
master ID: 723 (R=24 kOhm)
slave ID: 747 (R=27 kOhm)

- detecting HDI ID when detector server is up
- set master or slave (set "reg 0x20" bit-16 to 1 for slave)
- write the HDI module id to bit-17-19 in "reg 0x20"
- able to set master from server command line, server config file, client
- checkDetectorType signature changed in server (m3 and g2) to take error message pointer
- all virtual servers should also print the versions


erikfrojdh (approved):
LGTM

PR #518, opened 2022-08-15
Jungfraumaster

thattil:
- set master from client only
- also fixed (infinite recursion) bugs in ctb and moench server that are never called


erikfrojdh (approved):
This one as well!

PR #519, opened 2022-08-16
Jungfrausync

thattil:
jungfrau sync


python/src/detector.cpp
```diff
@@ -30,31 +30,31 @@ void init_det(py::module &m) {
         .def(py::init<int>())
 
         .def("freeSharedMemory",
-             (void (Detector::*)()) & Detector::freeSharedMemory)
+             (void(Detector::*)()) & Detector::freeSharedMemory)
```
erikfrojdh, line 33: On which PC did you run the generation scripts? Looks like clang-format removed a space. Maybe we can fix this to avoid a huge diff?

slsDetectorSoftware/src/Module.h
```diff
@@ -123,6 +123,8 @@ class Module : public virtual slsDetectorDefs {
     void setFlipRows(bool value);
     bool isMaster() const;
     void setMaster(const bool master);
+    bool getSynchronization() const;
+    void setSynchronization(const bool value);
```
erikfrojdh, line 127: const bool, is not having any effect in declaration only needed later in definition.

erikfrojdh (commented):
Looks good. I wonder if we can get around the white space diff. For the const we can also keep it and remove all of them at once.

slsDetectorSoftware/src/Module.h
thattil: but didnt we decide we will do this in a PR later where we remove all the const in declarations (except the ones with references)?

erikfrojdh: Makes sense!

python/src/detector.cpp
erikfrojdh: Pushed a new version without the white space diff.

erikfrojdh (approved):

PR #522, opened 2022-08-18
Nanosecond times in Python 

erikfrojdh:
As described in the bug Python datetime supports only micro seconds as lowest unit. To fix this we introduce a new class sls::Duration that wraps a uint64_t. The class is only used from Python

**Desired behavior**
Setting exposure time (and period, delay etc) is possible using either of:

- datetime.timedelat (Python standard library)
- seconds as float

Reading back times gives you either: 

- Time in seconds as float (d.exptime, d.period)
- Number of nanoseconds (equal to std::chrono::nanoseconds) with convenient conversion to seconds (d.getExptime(), direct map of C++ API) 

**Constraints**
No modifications to the C++ API

**Solution**

1. Introduce a new C++ type (DurationWrapper) which holds the number of nanoseconds as a uint64_t. This type is only used in the python bindings
2. remove the automatic chrono conversion in pybind11 (pybind11/chrono.h) not including the header is enough. 
3. add a new custom typecaster which does the conversion to and from std::chrono::nanoseconds

**sample code with PR**

```python
d = Detector()
d.exptime = 10e-9

>>> d.exptime 
1e-08

>>> d.getExptime()
[sls::DurationWrapper(total_seconds: 1e-08 count: 10)]

#Set using the standard library timedelta 
from datetime import timedelta
t = timedelta(microseconds = 3)
>>> d.exptime = t 

#Use the DurationWrapper, can be initialized with time in seconds
from slsdet import DurationWrapper
t = DurationWrapper(1.2)
d.exptime = t
>>> d.exptime

#Directly set the number of nanoseconds
t = DurationWrapper()
t.set_count(500)
>>> t
sls::DurationWrapper(total_seconds: 5e-07 count: 500)

```


thattil (approved):
LGTM

PR #583, opened 2022-11-21
Static internal build of qwt

erikfrojdh:
This PR includes a compressed qwt 6.1.5 that is unpacked and built as a static library. This allows us to remove qwt as an external dependency and reduces the risk of picking up the wrong version. 

1. Removed the qwt sources and docs
2. Created a new folder libs/qwt with a compressed qwt 6.1.5 and a CMakeLists file 
3. On configure cmake extracts the source code for qwt and builds a static library 
4. The static version of qwt is linked with our GUI

Clean builds on gcc12, gcc4.8 and clang 14 (other warnings in the slsDetectorPackage)


thattil (dismissed):

thattil (changes requested):
Please fix the warning for (qt5.15+) versions before merging. Otherwise LGTM!

erikfrojdh:
Pushed a fix for the warnings. Looks like we can ignore them and wait for qwt to fix it. 
https://doc.qt.io/qt-5/qflags-obsolete.html

PR #584, opened 2022-11-21
txdelay for all modules

thattil:
setting txdelay for all modules for eiger, jungfrau and m3


slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -431,6 +431,75 @@ void DetectorImpl::setGapPixelsinCallback(const bool enable) {
...
+        break;
+    default:
+        throw RuntimeError(
+            "Transmission delay is not implemented for the this detector.");
+    }
+    for (int i = 0; i != size(); ++i) {
+        if (eiger) {
+            modules[i]->setTransmissionDelayLeft(3 * i * step);
+            modules[i]->setTransmissionDelayRight((3 * i + 1) * step);
+            modules[i]->setTransmissionDelayFrame((3 * i + 2) * step);
```
erikfrojdh, line 496: In our old scripts we had a fixed txdelay_frame. I think we need it for multi frame acq but maybe clarify with Martin?

```bash
for i in `seq 0 $((NUM_MOD-1))`; do
	sls_detector_put $i:txndelay_left  $((i*2*DELAY_DIFF))
	sls_detector_put $i:txndelay_right $(((i*2+1)*DELAY_DIFF))
done
       sls_detector_put txndelay_frame  $((NUM_MOD*2*DELAY_DIFF))


```

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -431,6 +431,75 @@ void DetectorImpl::setGapPixelsinCallback(const bool enable) {
...
+    if (eiger) {
+        // between left and right
+        step = modules[0]->getTransmissionDelayRight();
+    } else {
+        // between first and second
+        step = modules[1]->getTransmissionDelayFrame();
+    }
+    for (int i = 0; i != size(); ++i) {
+        if (eiger) {
+            if ((modules[i]->getTransmissionDelayLeft() != (3 * i * step)) ||
```
erikfrojdh, line 461: Not 100% sure if this should be an error or -1 like with the threshold. Somehow both are ok.

erikfrojdh (changes requested):

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: Alright, so I will return -1 then

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: done

erikfrojdh (approved):

PR #591, opened 2022-12-09
udpate documentation on qt5 gui

thattil:


docs/src/installation.rst
```diff
@@ -206,6 +212,52 @@ using this compiler
...
+        #Add channels for dependencies and our library
+        conda config --add channels conda-forge
+        conda config --add channels slsdetectorgroup
+        conda config --set channel_priority strict
+
+        # create environment to compile
+        # on rhel7
+        conda create -n slsgui zeromq gxx_linux-64 gxx_linux-64 mesa-libgl-devel-cos6-x86_64 qt
+        # on fedora or newer systems
+        conda create -n slsgui zeromq qt
```
erikfrojdh, line 244: in case you use conda compilers you would also need the libgl, but no need for it on fedora unless maybe using it with ROOT

thattil: done

PR #595, opened 2022-12-12
cant stop

thattil:
fix to not throw exception when stopping detector when it really stopped, expecting idle instead of stopped flag


erikfrojdh:
Code looks good, merge if HW tests pass.

PR #602, opened 2022-12-14
eiger python versions fail

thattil:
fix for versions command in python failing for eiger due to hardware version not implemented


erikfrojdh (commented):
I wonder if it would be better to have hardware versions returning '1' for Eiger to avoid the added complexity here. How do you solve it in the command line?

thattil:
> I wonder if it would be better to have hardware versions returning '1' for Eiger to avoid the added complexity here. How do you solve it in the command line?

In the same way, not calling it for Eiger. Returning '1' for hardware version makes sense.

Just so you know though, in the developer version, we also have febl and febr firmware versions printed in the latest PR (which don't exist for the other detectors). So, we will still need to do this if logic.

So add hardware version for Eiger?

thattil:
> > I wonder if it would be better to have hardware versions returning '1' for Eiger to avoid the added complexity here. How do you solve it in the command line?
> 
> In the same way, not calling it for Eiger. Returning '1' for hardware version makes sense.
> 
> Just so you know though, in the developer version, we also have febl and febr firmware versions printed in the latest PR (which don't exist for the other detectors). So, we will still need to do this if logic.
> 
> So add hardware version for Eiger?

ok decided internally that we dont add this hardware version as it will be forever 1.

PR #604, opened 2022-12-19
ensuring no duplicate rx hostname port combo

thattil:
rx_hostname and port combo to one, or 
hostname to all, or 
a vector of hostnames and ports, 

- ignoring none or empty
- then verifying no duplicates for the host port combo including from shared memory


erikfrojdh:
Would it be possible to have the logic for verification of hostnames and port combinations in a free function? This way we could test that without having to have a (virtual) detector?

erikfrojdh (approved):

PR #609, opened 2023-01-16
fixed hostname not split (+) in python

erikfrojdh:
Since we do the splitting from string to vector of string in the CmdProxy it got omitted from the python code. This PR adds a fix to the python API. 

The parsing function is also factored out to allow for unit testing.


thattil (approved):
LGTM

PR #615, opened 2023-01-16
added missing equality operator for DurationWrapper in Python

erikfrojdh:
This restores the behavior for Mythen3 exposure times. Meaning as long as all the exposure times are the same the python interface returns a single value. 

The regression happened when moving form the python datetime.timedelta to our sls::DurationWrapper.


thattil (approved):
LGTM

PR #638, opened 2023-01-26
fix to access to shared memory that doesnt exist

thattil:


slsDetectorSoftware/src/SharedMemory.h
```diff
@@ -64,8 +64,23 @@ template <typename T> class SharedMemory {
             unmapSharedMemory();
     }
 
-    T *operator()() { return shared_struct; }
-    const T *operator()() const { return shared_struct; }
+    inline std::string getNoShmAccessMessage() const {
```
erikfrojdh, line 67: Could/should we make this a private function since the message is only used for the exception? I also think inline is not needed, usually the optimizer does it's thing anyway (even with inline it doesn't always get inlined)


```cpp
//alternative to avoid allocating the intermediate string. 
const char* noShmErrorMessage()
```

erikfrojdh (commented):
Small change in error message then merge?

slsDetectorSoftware/src/SharedMemory.h
thattil: makes sense. will do

PR #642, opened 2023-01-30
unicast udp_srcmac

thattil:
udp_srcmac can only be a unicast address (LSB of first octet must be 0)


slsSupportLib/include/sls/versionAPI.h
```diff
@@ -4,11 +4,11 @@
 #define RELEASE      "7.0.0.rc1"
```
erikfrojdh, line 4: ?

erikfrojdh (approved):

slsSupportLib/include/sls/versionAPI.h
thattil: Before it used to be called branch. In the end, its used in "versions" command as "packageversion"

PR #646, opened 2023-02-02
Fix rx arping socket bind

thattil:
- popen to execute arping was forking so it maintained that the socket was open (if it was forked when the udpsocket was open) when it was not anymore in a scropt of start and stop reciever. So Forking before calling popen and then moving delete udp pointer to after the listening threads are done running so that rx_arping cannot be set in between an acquiistion or when udp socket is already open.


slsReceiverSoftware/src/Listener.cpp
```diff
@@ -184,6 +184,14 @@ void Listener::ShutDownUDPSocket() {
     }
 }
 
+void Listener::DeleteUDPSocket() {
+    if (udpSocket) {
+        udpSocket = nullptr;
```
erikfrojdh, line 189: udpSocket.reset() might be more clear but would to the same thing.

erikfrojdh (commented):
Remove loop otherwise looks ok

slsReceiverSoftware/src/Arping.cpp
```diff
@@ -33,59 +35,68 @@ void Arping::SetInterfacesAndIps(const int index, const std::string &interface,
...
+        // just print (was already tested at Process start)
         if (!error.empty()) {
             LOG(logERROR) << error;
         }
 
-        // wait for 60s as long as thread not killed
+        // wait for 60s as long as Process not killed
         int nsecs = 0;
-        while (runningFlag && nsecs != 60) {
+        while (nsecs != timeIntervalSeconds) {
```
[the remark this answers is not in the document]
thattil: ah yes, not anymore. will fix

slsReceiverSoftware/src/Listener.cpp
thattil: done

PR #659, opened 2023-02-14
fixed linking issue in arping and failing shm test

erikfrojdh:
Addresses two issues:

- one of the shm tests were failing since we throw on trying to get shm when it holds a nullptr. --> updated test
- linking issue in debug build of arping --> moved unnecessary static member to a local variable


thattil (approved):
LGTM

PR #690, opened 2023-03-15
ctb slow adc

thattil:
- reduced usleep to us precision
- trailing edge for reading data
- clk kept low after setting register


erikfrojdh (approved):

thattil:
Already resolved with #713  and new fimware. This PR is discarded.

PR #691, opened 2023-03-15
rx_arping handle sigchld

thattil:
rx_arping pclose gave -1 due to sigchld being ignored, fixed with sighandler doing a wait


erikfrojdh (approved):
LGTM, have some unrelated changes but we can look into them later.

PR #697, opened 2023-03-17
fix old server version in 64 bits

thattil:
check server version before initial checks, catch old server version exception, get old server version as 64 bit and print it along with exception


erikfrojdh (approved):
LGTM, best would of course be to also have a test for this...

PR #703, opened 2023-03-24
Fix ctb slow adc datasheet

thattil:


thattil:
Already resolved with #713  and new fimware. This PR is discarded.

PR #717, opened 2023-04-12
1 - Add cmake for jungfrauExecutables rawdataprocess

vhinger182:


erikfrojdh (commented):
Looks good to me. I would also remove the Makefiles at the same time so we only have one way of building the zmq processes otherwise the risk is very high that one of them is out of sync. 

What would be good next time is to have only one change in the PR, now there are some small code changes in addition to the CMake stuff...

vhinger182:
> Looks good to me. I would also remove the Makefiles at the same time so we only have one way of building the zmq processes otherwise the risk is very high that one of them is out of sync.
> 
> What would be good next time is to have only one change in the PR, now there are some small code changes in addition to the CMake stuff...

I will talk to Anna and figure out if we can delete the Makefiles. Because, for now, not all Makefiles are implemented in CMake and not all objects in the Makefile are valid (we think).

erikfrojdh:
Sounds like I plan. I will merge this and then step by step we try to remove the Makefiles.

erikfrojdh (approved):

PR #719, opened 2023-04-20
7.0.2 changes to developer

thattil:
- Merging 7.0.2.rc to developer
- also deleting 7.0.0 binaries


slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1242,25 +1242,45 @@ int DetectorImpl::acquire() {
...
-void DetectorImpl::startAcquisition(bool blocking, std::vector<int> positions) {
-    // handle Mythen3 synchronization
-    if (shm()->detType == defs::MYTHEN3 && size() > 1) {
-        std::vector<int> master;
-        std::vector<int> slaves;
-        if (positions.empty() ||
-            (positions.size() == 1 && positions[0] == -1)) {
-            positions.resize(modules.size());
-            std::iota(begin(positions), end(positions), 0);
+void DetectorImpl::getMasterSlaveList(std::vector<int> positions,
```
erikfrojdh, line 1245: Maybe better to use const& like in the other places? 
```
//Easiest done with the using declaration we have in sls_detector_defs
using Positions = const std::vector<int> &;
```

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1281,6 +1303,58 @@ void DetectorImpl::startAcquisition(bool blocking, std::vector<int> positions) {
...
+        if (!master.empty())
+            Parallel(&Module::sendSoftwareTrigger, master, block);
+    }
+    // all in parallel
+    else {
+        Parallel(&Module::sendSoftwareTrigger, positions, block);
+    }
+}
+
+void DetectorImpl::stopDetector(std::vector<int> positions) {
```
erikfrojdh, line 1332: stop only has special case for Jungfrau not M3?

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1281,6 +1303,58 @@ void DetectorImpl::startAcquisition(bool blocking, std::vector<int> positions) {
...
+
+    bool handleSynchronization = false;
```
erikfrojdh, line 1334: Could we make a function for the handle synchronization logic? I see it in two places here. Risk is that we only change one if we upgrade =)

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1242,25 +1242,45 @@ int DetectorImpl::acquire() {
...
+        else
+            slaves.push_back(i);
+    }
+}
+
+void DetectorImpl::startAcquisition(const bool blocking,
+                                    std::vector<int> positions) {
+    bool handleSynchronization = false;
+    // multi module m3 or multi module sync enabled jungfrau
+    if (size() > 1) {
```
erikfrojdh, line 1268: Is there a way around having to differentiate between sometimes M3 sometimes not?

erikfrojdh (changes requested):

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: I wanted to do that as well, but they are quite different for each of the functions. For the software trigger, its not for M3. For the stop, its slaves first. Or is there another way to handle this? I couldnt find a way without cluttering.

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: Only the software trigger is not implemented for M3.,. so..its either multi module m3 or its not..
but we also check for multi module jungfrau that has sync enabled..

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: This is the only place where we modify positions, but will do it in the other places where we dont.

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: Ill add M3 this time and G2 and G1 in another PR after what the next meeting entails.

thattil: done

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1242,25 +1242,45 @@ int DetectorImpl::acquire() {
...
+    auto is_master = Parallel(&Module::isMaster, positions);
+    for (size_t i : positions) {
+        if (is_master[i])
+            master.push_back(i);
+        else
+            slaves.push_back(i);
+    }
+}
+
+void DetectorImpl::startAcquisition(const bool blocking,
```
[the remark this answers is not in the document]
thattil: done

thattil:
Aborting this PR. Mergin 7.0.2.rc directly

PR #721, opened 2023-04-20
7.0.2.rc

thattil:
- row and col for multi mod junngfrau single interface (gui)
    - jungfrau multi mod sync mode (start, trigger, stop)
    - set/get timing jungfrau does not give error when syc enabled
    - switching between 1 and 2 interfaces did not set gui/client zmq port properly. Resulted in dummy streaming forever.+
    - adding missing cstdint for gcc 13


slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1256,25 +1256,44 @@ int DetectorImpl::acquire() {
...
-                slaves.push_back(i);
+        if ((shm()->detType == defs::JUNGFRAU) &&
+            (Parallel(&Module::getSynchronizationFromStopServer, pos)
+                 .tsquash("Inconsistent synchronization among modules"))) {
+            handleSynchronization = true;
         }
+    }
+
+    // slaves first
+    if (handleSynchronization) {
```
erikfrojdh, line 1293: I think handleSynchronization could be factored out as a function since the check is the same every time.

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1286,15 +1305,73 @@ void DetectorImpl::startAcquisition(bool blocking, std::vector<int> positions) {
...
+    bool handleSynchronization = false;
+    // multi module sync enabled jungfrau
+    if (size() > 1 && (shm()->detType == defs::JUNGFRAU) &&
+        (Parallel(&Module::getSynchronizationFromStopServer, pos)
+             .tsquash("Inconsistent synchronization among modules"))) {
+        handleSynchronization = true;
+    }
+
+    // slaves first
+    if (handleSynchronization) {
```
erikfrojdh, line 1329: How does Junfrau/Mythen3 work? I think we only send software trigger to master (or all if non sync)

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1286,15 +1305,73 @@ void DetectorImpl::startAcquisition(bool blocking, std::vector<int> positions) {
...
+        if (!master.empty())
+            Parallel(&Module::sendSoftwareTrigger, master, block);
+    }
+    // all in parallel
+    else {
+        Parallel(&Module::sendSoftwareTrigger, pos, block);
+    }
+}
+
+void DetectorImpl::stopDetector(Positions pos) {
```
erikfrojdh, line 1344: Since we anyway want to stop does the order actually matters?

erikfrojdh (changes requested):

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: Apparently it does, both @mozzanica and @anberga said so.

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: Yes. Will do

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: From yesterdays meeting, JF must be slaves first, then master. For M3: it doesnt matter, so I will do all of them like JF. G2/G1 dont have it implemented, but that as you said can give an exception if used from the detetor server itself.

thattil: Refactored software trigger for m3 as well.

slsDetectorSoftware/src/DetectorImpl.cpp
erikfrojdh: Looks good. Maybe in the future if we end up with many calls like this we can add a SynchronisedParallel call.

erikfrojdh (approved):
Looks good. Less special cases now!

PR #759, opened 2023-06-02
Ctb sense power signal names

thattil:
- added signal, power, sense names for ctb


python/slsdet/detector.py
```diff
@@ -1766,7 +1766,7 @@ def slowadc(self):
     @property
     def daclist(self):
         """
-        List of enums/names for every dac for this detector.
+        List of enums/names for every dac for this detector. 18 dacs
```
erikfrojdh, line 1769: Do all detectors have 18 dacs? Comment misleading?

thattil: Good catch! Will fix

erikfrojdh (approved):

PR #779, opened 2023-07-13
2. Set row col

thattil:
set custom row and column in the udp header. 

based on ctb_trasnsceiver_ro PR (steps: merge that, developer to this PR, visually easier PR then)


slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -2605,8 +2605,8 @@ int readSample(int ns) {
             // if channel is in enable mask
             if ((1 << ich) & (transceiverMask)) {
 
-                int offset = FIFO_TIN_STATUS_FIFO_EMPTY_1_OFST + ich;
-                uint32_t mask = (1 << offset);
+                // int offset = FIFO_TIN_STATUS_FIFO_EMPTY_1_OFST + ich;
```
erikfrojdh, line 2608: What is happening here?

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -10581,4 +10586,128 @@ int get_transceiver_enable(int file_des) {
...
+    return Server_SendResult(file_des, INT32, NULL, 0);
+}
+
+int get_column(int file_des) {
+    ret = OK;
+    memset(mess, 0, sizeof(mess));
+    int retval = -1;
+
+    LOG(logDEBUG1, ("Getting column\n"));
+    // get only
```
erikfrojdh, line 10644: Do we need the comment? Get should only do get... I hope?

erikfrojdh (approved):

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: every transceiver channel (4 in total) have a bit to represent if its fifo is empty or not.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
thattil: there are around 90 such comments and more that come from the time we rewrote the server client interface from scratch. it was more significant at the time. Maybe another PR to get rid of all these unnecessary commetns.

PR #780, opened 2023-07-14
Internal zmq using FetchContent

erikfrojdh:
- Adding internal version of zmq using FetchContent
- tar file in repo and option to pull from github


CMakeLists.txt
```diff
@@ -1,15 +1,62 @@
...
 include(cmake/project_version.cmake)
 include(cmake/SlsAddFlag.cmake)
-include(cmake/SlsFindZeroMQ.cmake)
+
+
+
+# Using FetchContent to get libzmq
+include(FetchContent)
+
+option(SLS_FETCH_ZMQ_FROM_GITHUB "HDF5 File format" OFF)
```
thattil, line 22: HDF5 File format?

thattil (changes requested):

CMakeLists.txt
erikfrojdh: Good catch. Fixed it =)

thattil (approved):

PR #803, opened 2023-09-15
commands code generation 

Bechir-Brahem:
usage: gen_commands.py [-h] [-f] [-p] [-c] [-g] [-a]

generate c++ code for cli commands from the commands.yaml file

options:
  -h, --help          show this help message and exit
  -f, --format        format header and cpp file using clang-format
  -p, --parse         parse the commands.yaml file into extended_commands.yaml
  -c, --check         check missing commands
  -g, --generate      generate code (C++ or bash if -a is used)
  -a, --autocomplete  print bash autocomplete values

```bash
# parse commands.yaml, generate c++, format c++ files
python gen_commands.py -pgf 

# get enum values from autocomplete/dump.json, and generate bash and zsh autocompletion scripts
python gen_commands.py -ag 
```

the `autocomplete/autocomplete.py` script reads the values associated with each `defs::*` enum found in `ToString.cpp` and prints them.
example of what it extracts from the cpp file:
```
defs::vetoAlgorithm
ALG_HITS "hits"
ALG_RAW "raw"
```
to do that it iterates over the AST of that file (found in dump.json)  
the command to generate the AST can be found in the comments of autocomplete.py file


slsDetectorSoftware/generator/commands.yaml
```diff
@@ -0,0 +1,19 @@
+---
+frames:
```
erikfrojdh, line 2: Looks clean! I like it, lets see if it works for more complex commands.

slsDetectorSoftware/src/CmdApp.cpp
```diff
@@ -9,15 +9,16 @@ int main(int argc, char *argv[]){
...
-    std::cout << "cmd: " << parser.command() << "\narguments: ";
-    for (auto& arg : parser.arguments()){
-        std::cout << arg << " ";
-    }
-    std::cout << "\n\n";
     
     int default_action = -1;
     sls::Detector d(parser.multi_id());
     sls::Caller c(&d);
+
```
erikfrojdh, line 16: This now calls both get and put which might be convenient for testing but has the problem that we need different number of arguments

erikfrojdh (changes requested):
Good start but a few comments: 

- Could you write the Caller.h/Caller.cpp to slsDetectorSoftware/src? Then we can build directly
- Add check for number of commands, now it segmentation faults if called with to few parameters

Once merged to the generator branch we can add the help string and some tests

PR #847, opened 2023-11-02
7.0.3.rc: gui acq finished callback for different status

thattil:
fix acquisition finished status to have different status for different modules, but does not have to be error. for eg. jf sync fw (2.4.1 gives idle for master and stopped for slaves when stopping acquiistion)


erikfrojdh (approved):
LGTM

PR #862, opened 2023-11-09
trigger signal issues handled at acquire

thattil:
Case: trigger signal has an issue (termination board or flat band cable not connected properly)

- if blocking and handling sync, only master gets blocking acq, slaves get non blocking as they are first and so dont get status checked. So no error thrown when slaves dont get trigger (due to not connected etc) and acq returns with slaves still in waiting status. so check status of all in blocking acq

- ensure a sync multi module (jf, m3, g2) has atleast one master when starting acquisition.

### DIFF
--- a/slsDetectorSoftware/src/DetectorImpl.cpp
+++ b/slsDetectorSoftware/src/DetectorImpl.cpp
@@ -1316,19 +1316,17 @@ void DetectorImpl::startAcquisition(const bool blocking, Positions pos) {
         if (!slaves.empty()) {
             Parallel(&Module::startAcquisition, slaves);
         }
-        if (!masters.empty()) {
-            if (blocking) {
-                Parallel(&Module::startAndReadAll, masters);
-                // ensure all status normal (slaves not blocking)
-                // to catch those slaves that are still 'waiting'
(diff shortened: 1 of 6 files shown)


slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1309,13 +1309,26 @@ void DetectorImpl::startAcquisition(const bool blocking, Positions pos) {
...
         std::vector<int> slaves;
         getMasterSlaveList(pos, masters, slaves);
+        if (masters.empty()) {
+            throw RuntimeError("Cannot start acquisition in sync mode. No "
+                               "master module found");
+        }
         if (!slaves.empty()) {
             Parallel(&Module::startAcquisition, slaves);
         }
         if (!masters.empty()) {
```
erikfrojdh, line 1319: I guess the 
```
if (!masters.empty())
```
is now redundant since we throw if it it empty?

erikfrojdh (commented):
Think you can remove one if, otherwise it looks good to me!

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: done

PR #869, opened 2023-11-16
forward declare zmq_msg_t and moved include

erikfrojdh:
Addressing error when building without zmq-devel installed.


thattil (approved):
LGTM

PR #873, opened 2023-12-01
Detector specific autocomplete

erikfrojdh:
Added detector specific autocomplete for 

1. dac
2. settings
3. timing

Based on the branch with deprecated commands


thattil (approved):
LGTM

PR #880, opened 2023-12-12
More helpful error messages for unsupported actions

erikfrojdh:
Modified codegen to allow for missing function and throw on true.


slsDetectorSoftware/generator/commands_parser/commands_parser.py
```diff
@@ -40,10 +40,13 @@ def __init__(
             'actions': {}
         }
 
-    def _verify_argument(self, arg, infer_action):
+    def _verify_argument(self, arg, infer_action, command_name, action):
         if arg['function'] == '' and 'ctb_output_list' not in arg:
-            #TODO can we get the name of the top level command?
-            print(f"Warning {arg} does not have a function")
+            special_exception_message_list = ["Cannot put", "Cannot get"]
+            if 'exceptions' in arg and any(ele in arg['exceptions'][0]['message'] for ele in special_exception_message_list):
```
erikfrojdh, line 46: Need to check for conditions and well to make sure that we get an if (true)

thattil: done

PR #907, opened 2024-02-05
Protect from getenv("HOME") returning nullptr 

fleon-psi:
* checks if getenv("HOME") is nullptr
* assert in safe_strcpy pointer is not null

### DIFF
--- a/slsDetectorSoftware/src/Module.cpp
+++ b/slsDetectorSoftware/src/Module.cpp
@@ -3332,7 +3332,13 @@ void Module::initializeModuleStructure(detectorType type) {
     shm()->numberOfModule.y = 0;
     shm()->controlPort = DEFAULT_TCP_CNTRL_PORTNO;
     shm()->stopPort = DEFAULT_TCP_STOP_PORTNO;
-    strcpy_safe(shm()->settingsDir, getenv("HOME"));
+    char *home_directory = getenv("HOME");
+    if (home_directory != nullptr)
+        strcpy_safe(shm()->settingsDir, home_directory);
+    else {
+        strcpy_safe(shm()->settingsDir, "");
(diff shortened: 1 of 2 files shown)


slsDetectorSoftware/src/Module.cpp
```diff
@@ -3332,7 +3332,13 @@ void Module::initializeModuleStructure(detectorType type) {
     shm()->numberOfModule.y = 0;
     shm()->controlPort = DEFAULT_TCP_CNTRL_PORTNO;
     shm()->stopPort = DEFAULT_TCP_STOP_PORTNO;
-    strcpy_safe(shm()->settingsDir, getenv("HOME"));
+    char *home_directory = getenv("HOME");
+    if (home_directory != nullptr)
+        strcpy_safe(shm()->settingsDir, home_directory);
+    else {
+        strcpy_safe(shm()->settingsDir, "");
+        std::cout << "Warning, HOME directory not set" << std::endl;
```
thattil, line 3340: Could you replace this with "LOG(logWARNING) << "HOME directory not set";

So that it prints in yellow.

thattil (changes requested):
Could you please modify the print to what is suggested? I would have done it myself but I cant seem to manage to access this. Thank you so much for finding this fix! :)

After we merge this, I will cherry pick this commit and merge to the developer as well.

slsDetectorSoftware/src/Module.cpp
fleon-psi: @thattil please try if this is OK now.
I have no idea how to set the proper permission, in the pull request options "allow edits by maintainers" is set, but I'm no expert in Github.

thattil: @fleon-psi All good. Ill merge it now :)

PR #926, opened 2024-07-16
developer: blocking acquire stop with slave temporarily in waiting

thattil:
client: stopping a blocking acquire of multi modules checks status to catch slaves that might still be in waiting.

Problem is (gotthard2 25um at least) slave is in waiting only temporarily before going go idle/stopped.

So a 50ms sleep is necessary ot not throw an unnecessary error


erikfrojdh (approved):
Looks good. Changed the title to reflect the branch we are merging to

PR #931, opened 2024-07-26
bug fixes related to file saving

ZhenghengLi:
* fix the file path resetting issue of GUI in case where different modules have different fpath settings.
* fix the stack-buffer-overflow issue when using HDF5 file format

### DIFF
--- a/slsReceiverSoftware/src/MasterFileUtility.cpp
+++ b/slsReceiverSoftware/src/MasterFileUtility.cpp
@@ -215,7 +215,7 @@ std::string CreateVirtualHDF5File(
 
         // property list
         H5::DSetCreatPropList plist;
-        long fill_value = -1;
+        uint64_t fill_value = -1;
         plist.setFillValue(dataType, &fill_value);
         std::vector<H5::DSetCreatPropList> plistPara(paraSize);
         // ignoring last fill (string)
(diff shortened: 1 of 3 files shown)


erikfrojdh (approved):

slsReceiverSoftware/src/MasterFileUtility.cpp
```diff
@@ -215,7 +215,7 @@ std::string CreateVirtualHDF5File(
 
         // property list
         H5::DSetCreatPropList plist;
-        int fill_value = -1;
+        long fill_value = -1;
```
thattil, line 218: Could you please change this to a uint64_t? Could you also change it in HDF5DataFile.cpp: there is also a variable 'fill_value' that is of datatype 'int' and that should be changed to 'uint64_t'?

thattil (changes requested):
Sorry for the delay in getting back to you. Thank you for finding and resolving these bugs! We appreciate them :)

We just have a minor change request before we can merge it into the 8.0.2 release candidate.

slsReceiverSoftware/src/MasterFileUtility.cpp
ZhenghengLi: Yes, uint64_t is better. As to HDF5DataFile.cpp, I see the size of dataType in the next line (i.e. line 190) is at most 32bit, thus didn't do the change before, but using uint64_t is safer anyway. I have changed the type of fill_value to uint64_t in both files.

thattil (approved):

PR #933, opened 2024-08-05
m3: fixed stop server not starting up with setup variables

thattil:


thattil:
replaced by another PR with branch dev/virtual_stop_server_start

PR #958, opened 2024-09-04
Dev/zmq stream all ipv6 adn remove rx_zmqip

thattil:
- 'Depreciated' changed to 'deprecated'
- removed rx_zmqip removed from python, API 
- kept rx_zmqip command line (only in CallerSpecial to give warning)
- ZmqSocket publisher socket constructor does not take in IP anymore, uses hardcoded '0.0.0.0' instead (used in Datastreamer, in MoenchZmqProcess and slsSupportLib/tests for zmqSocket
- gui: just a read only that shows "0.0.0.0", removed signals and slots for rx_zmqip
- client: changed updateReceiverStreamingIp to updateClientStreamingIP so that the zmqip is updated with rx_hostname by default if empty
- -zmqsocket: added one more error print (ENOENT no endpoint)


slsSupportLib/include/sls/sls_detector_funcs.h
```diff
@@ -748,8 +748,8 @@ const char* getFunctionNameFromEnum(enum detFuncs func) {
 	case F_GET_RECEIVER_FILE_FORMAT:		return "F_GET_RECEIVER_FILE_FORMAT";
 	case F_SET_RECEIVER_STREAMING_PORT: 	return "F_SET_RECEIVER_STREAMING_PORT";
 	case F_GET_RECEIVER_STREAMING_PORT: 	return "F_GET_RECEIVER_STREAMING_PORT";
-	case F_SET_RECEIVER_STREAMING_SRC_IP: 	return "F_SET_RECEIVER_STREAMING_SRC_IP";
-	case F_GET_RECEIVER_STREAMING_SRC_IP: 	return "F_GET_RECEIVER_STREAMING_SRC_IP";
+	case F_SET_RECEIVER_STREAMING_SRC_IP: 	return "F_SET_RECEIVER_STREAMING_SRC_IP - obsolete";
+	case F_GET_RECEIVER_STREAMING_SRC_IP: 	return "F_GET_RECEIVER_STREAMING_SRC_IP - obsolete";
```
erikfrojdh, line 752: Are we using this in another place? Where is obsolete defined?

erikfrojdh (approved):
Looks good just one small question.

PR #966, opened 2024-09-11
Dev/rx callbacks

thattil:
-  formatted the pyctbgui/src c files
- formatted others in server side
- changed rxr callback signatures to all include structs
- bringing some parameters (set functions) to dataProcessor class for its callback (namely udpport, quad, fliprows, totalframes, jsonheader), resulting in also removing totalframes from 2 other function signatures
- updated MultiReceiverApp to reflect the new callback signatures


felix-engelmann:
As we normally store stuff in the addJsonHeader which is required on arm, it would be nice to add the addJsonHeader to the start acquisition callback.

thattil:
> As we normally store stuff in the addJsonHeader which is required on arm, it would be nice to add the addJsonHeader to the start acquisition callback.
And keep the one in the datacallback?

felix-engelmann:
> And keep the one in the datacallback?

yes, as you mentioned that the additional values can change during the scan.

thattil:
> > And keep the one in the datacallback?
> 
> yes, as you mentioned that the additional values can change during the scan.

done

slsSupportLib/include/sls/sls_detector_defs.h
```diff
@@ -160,6 +176,36 @@ class slsDetectorDefs {
         sls_detector_header detHeader; /**< is the detector header */
         sls_bitset packetsMask;        /**< is the packets caught bit mask */
     };
+
+    struct startCallbackHeader {
+        std::vector<uint32_t> udpPort;
```
erikfrojdh, line 181: How many ports can we have?

erikfrojdh (commented):
My main question at this moment is what the user will do with the headers we supply. Do they need to keep them longer than the callback? Do we get RVO?

slsSupportLib/include/sls/sls_detector_defs.h
thattil: 1 or 2 udp ports per startcallback and endcallbackheader

erikfrojdh: I'm torn if we should keep a vector that has one or two elements or if we should use our own StaticVector to avoid heap allocations.

PR #968, opened 2024-09-11
Dev/frame synchronizer

thattil:
@felix-engelmann
Please make your PRs to this branch


felix-engelmann:
Hi @thattil Thanks for cleaning the branch. Is there anything missing before a merge? And are the callbacks stable now?

thattil:
> Hi @thattil Thanks for cleaning the branch. Is there anything missing before a merge? And are the callbacks stable now?

Hi @felix-engelmann, thanks for developing all this code :) Were the callbacks not stable, they look quite stable now. Is there something to look out for?

Just review left to merge.

erikfrojdh (approved):

thattil:
> IIRC, there was a question if the additional JSON is present on every frame, such that if they are updated during an acquisition, the values in the stream headers change. And are we happy using the UDP port to uniquely identify half modules? Or is there a better way? And that led to the variable length vector, which is only ever 1 or 2.

Yes,  json header could change for every frame depending on the user. So, I have added it back. Identifying the half modules/modules is best with udp port at the moment. Modid would have been another option, but this is sometimes not connected for some detectors in the firmware yet. we can always modify this if we think of something else. 

We did find some issues with the sanitizer. Was to do with freeing instead of deleting and not deleting zmq_msg_t messages. I have made the fixes in the PR #1086 . Once this is reviewed, I will port the changes here.

PR #990, opened 2024-09-30
python version and versions command

thattil:
fixing python version. allowing versions to execute even if no detector connected or in shared memory


python/slsdet/__init__.py
```diff
@@ -27,4 +27,14 @@
 scanParameters = _slsdet.scanParameters
 currentSrcParameters = _slsdet.currentSrcParameters
 DurationWrapper = _slsdet.DurationWrapper
-pedestalParameters = _slsdet.pedestalParameters
\ No newline at end of file
+pedestalParameters = _slsdet.pedestalParameters
+
+
+import subprocess
+def get_git_tag():
```
erikfrojdh, line 34: I don't think this works well since the conda pkg will never have access to the git tree

PR #1002, opened 2024-10-18
900/proper free

thattil:
- removed class member function freeSharedmemory for both Detector and Module
- made the free function freeSharedmemory accessible to python interface
- setHostname if there is already a module in shm will recreate the Detector object while freeing shm completely and keeping detsize and intitialchecks
- sethostname called from DetectorClass in virtual command to have one point of entry
- testing Module class frees shared memory using free function
- Detector class: added copy and move constructor and assignmentoperators due to explicit destructor (DetectorImpl fwd declared)
- DetectorImpl class: included ZmqSocket to remove destructor (should not be virtual in any case)
- Module class: removed explciit destructor to allow compiler generated constructor and operators

### DIFF
--- a/slsDetectorSoftware/src/DetectorImpl.h
+++ b/slsDetectorSoftware/src/DetectorImpl.h
@@ -79,11 +79,7 @@ class DetectorImpl : public virtual slsDetectorDefs {
     explicit DetectorImpl(int detector_index = 0, bool verify = true,
                           bool update = true);
 
-    /**
-     * Destructor
-     */
     virtual ~DetectorImpl();
-
     template <class CT> struct NonDeduced { using type = CT; };
     template <typename RT, typename... CT>
     Result<RT> Parallel(RT (Module::*somefunc)(CT...),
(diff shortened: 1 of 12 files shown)


slsDetectorSoftware/src/DetectorImpl.h
```diff
@@ -5,6 +5,7 @@
 #include "CtbConfig.h"
 #include "SharedMemory.h"
 #include "sls/Result.h"
+#include "sls/ZmqSocket.h"
```
erikfrojdh, line 8: Is it not possible to forward declare this anymore?

slsDetectorSoftware/src/Module.h
```diff
@@ -75,12 +75,6 @@ class Module : public virtual slsDetectorDefs {
     verify is if shared memory version matches existing one */
     explicit Module(int det_id = 0, int module_index = 0, bool verify = true);
 
-    virtual ~Module();
```
erikfrojdh, line 78: Need to double check this...

erikfrojdh (commented):
Will check some thing and then we can bring it in.

slsDetectorSoftware/src/DetectorImpl.h
erikfrojdh: No issue since it's in DetectorImpl and not in the public header

PR #1017, opened 2024-10-31
900/detach pthreads not joining

thattil:
In the ctb server for 1g non blocking acquisitions:
- the thread to read fifo exits on its own needs to be detached if not joined. So memory leak. Fixed
- possibly also affects simulators sending data


slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
```diff
@@ -2572,6 +2572,9 @@ int startStateMachine() {
         sharedMemory_setStatus(IDLE);
         sharedMemory_unlockAcqFlag();
         return FAIL;
+    } else {
+        // clean up
```
erikfrojdh, line 2576: Clean up?

erikfrojdh (approved):
Looks good to me. Maybe remove the //clean up comments?

erikfrojdh (approved):

PR #1019, opened 2024-10-31
Dev/detach pthreads not joining

thattil:
In the ctb server for 1g non blocking acquisitions:
- the thread to read fifo exits on its own needs to be detached if not joined. So memory leak. Fixed
- possibly also affects simulators sending data


erikfrojdh (approved):
Same as for 9.0.0. Do we need the comments?

PR #1024, opened 2024-11-08
9.0.0: version from file to cmake and python

thattil:
- renamed conda-recipe folder
- added a check to see if build and install folder exists in build.sh (conda recipe)
- created VERSION file that has '0.0.0'for developer but can be updated using update_version.py that takes in a version. The script checks for semantic versioning and updates VERSION file
- VERSION file also copied along with py files to slsdet in python cmakelist and build_pylib.sh (for conda), also copied in root folder for installations (for no coding purpose)
- init.py and setup.py reads this file to get the version (a bit differently to find the VERSION file)
- VERSION file read into cmake to get the version and also added to compile definition. So RELEASE removed from versionAPI.h (using SLS_DET_VERSION compile definiton instead) and also removed updateRelease script.
- conda getting project version from environment variable SLS_DET_VERSION that is set in build_pylib.sh prior.
- added 3.13 python to conda build
- anything related to ctb removed from release notes as users will always use developer
- sets 9.0.0 to VERSION file  by running update_version.py without an argument


erikfrojdh (commented):
Did you test this with the conda packages?

PR #1064, opened 2024-11-28
ctb:change exptime if runclk changes

thattil:


thattil:
changed to https://github.com/slsdetectorgroup/slsDetectorPackage/issues/1072

PR #1086, opened 2025-02-04
910/frame synchronizer

thattil:


slsReceiverSoftware/src/FrameSynchronizerApp.cpp
```diff
@@ -0,0 +1,575 @@
...
+        all_keys; // Set to collect all unique keys across all maps
+    std::set<long unsigned int> valid_keys; // Set to store final valid keys
+
+    // If no maps are provided, return empty set
+    if (maps.empty()) {
+        return valid_keys;
+    }
+
+    // Collect all unique keys from all maps
+    for (const auto &[port, trigger_map] : maps) {
```
erikfrojdh, line 115: Structured bindings requires c++17. 

This would break compilation on RH7 (gcc4.8). Which is ok but I guess we need to inform the group. This definitely makes in a major version.

slsReceiverSoftware/src/FrameSynchronizerApp.cpp
```diff
@@ -0,0 +1,575 @@
...
+
+/** Structure handling different threads */
+struct Status {
+    bool starting = true;
+    bool terminate = false;
+    unsigned long num_receivers;
+    sem_t available;
+    std::mutex mtx;
+    std::vector<zmq_msg_t *> headers;
+    std::map<unsigned int,
```
erikfrojdh, line 47: De facto, it's the same type but I would be in favour of using uint64_t since that is what we have in the header.

slsReceiverSoftware/src/FrameSynchronizerApp.cpp
```diff
@@ -0,0 +1,575 @@
...
+    struct sigaction asa;
+    asa.sa_flags = 0;          // no flags
+    asa.sa_handler = SIG_IGN;  // handler function
+    sigemptyset(&asa.sa_mask); // dont block additional signals during
+                               // invocation of handler
+    if (sigaction(SIGPIPE, &asa, nullptr) == -1) {
+        cprintf(RED, "Could not set handler function for SIGPIPE\n");
+    }
+
+    Status stat{true, false, (unsigned long)numReceivers};
```
erikfrojdh, line 538: Maybe we can use static_cast, or make it the correct type from the beginning? size_t would make sense right?

erikfrojdh (changes requested):
Minor changes. But needs address and thread sanitizers to pass.

slsReceiverSoftware/src/FrameSynchronizerApp.cpp
thattil: Yes. Agreed

erikfrojdh (commented):
Looks good. Could maybe be cleaner with range based for loops. Need to test with a range of detectors. C++17 would be nice. auto [key, value]

slsReceiverSoftware/src/FrameSynchronizerApp.cpp
```diff
@@ -0,0 +1,622 @@
...
+std::set<uint64_t> get_valid_fnums(const PortFrameMap &port_frame_map) {
+    // empty list
+    std::set<uint64_t> valid_fnums;
+    if (port_frame_map.empty()) {
+        return valid_fnums;
+    }
+
+    // collect all unique frame numbers from all ports
+    std::set<uint64_t> unique_fnums;
+    for (auto it = port_frame_map.begin(); it != port_frame_map.begin(); ++it) {
```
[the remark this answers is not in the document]
thattil: Yes, we cannot do upgrade to c++17 in a minor version in any case.

PR #1114, opened 2025-02-21
910: multi ids ignored silently in config file

thattil:


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -97,8 +97,15 @@ void Detector::loadParameters(const std::string &fname) {
 void Detector::loadParameters(const std::vector<std::string> &parameters) {
     Caller caller(this);
     CmdParser parser;
+    const int detShmId = getShmId();
     for (const auto &current_line : parameters) {
         parser.Parse(current_line);
+        int m_id = parser.multi_id();
+        if (m_id != detShmId) {
```
erikfrojdh, line 104: Is this the check we want to have? Right now you have to specify the multi_id in the config file

```bash
p config 1-mydet.conf #doesn't work unless all commands have 1- in the config file
```

I guess the correct check is 
```cpp
if (parser.multi_id()!=0)
   throw...
```

erikfrojdh (changes requested):

slsDetectorSoftware/src/Detector.cpp
thattil: I agree. It should throw no matter what multi id is given in the config file.

erikfrojdh (approved):

PR #1149, opened 2025-03-14
update ctb regDefs, included fill level of adc, transceiver and DBit, cont. readout enable regs

### LINKED ISSUE #1144: ctb: check half empty instead of not empty flag before reading n samples (a frame)
Assuming the user will set  4096 (half fifo or 80%) number of samples, the ctb checks a bit that only looks for half empty (or 80%) flag before reading n samples.

@lopez-c @siebsi 
If its not a lot of work, instead of this idea, if the fifo reaches a certain "x" number of samples  (which is set via a register by the user), the firmware will set a bit is set in some register, which the ctb server can poll (instead of not 'empty' bit) to start reading "n" samples (total number of samples in a frame). This way, there is flexibility for the user to test. 

Cheers,
Dhanya

mmarti04:
- Removed deprecated Register definitions
- added Registers for FiFo Fill levels and enable of continuous readout

When continuous ro for adc is enabled --> packets are send whenever ANY of the adc FiFos reaches 75% Full. Same for Transceiver and DBIT. 

Do we need this enable per FiFo / 75% limit adjustable ? Enable per FiFo would be easy, adjustable limit will likely get me in trouble with  timing. 

@thattil: FIFO_TIN_FILL_REG and FIFO_DIN_STATUS_FIFO_FILL_OFST gives you the fill level of the transceiver and DBit fifo at any time. The ADC FiFo fill level is muxed with the same FIFO_CHNNL_SLCT that is used for 1G readout of the ADC. So you have to write this and then you can read the corresponding fill level at FIFO_ADC_FILL_REG

(Fill level = Number of words in the fifo)


erikfrojdh:
@thattil  Do we need to update the FW version checks?

thattil:
> Do we need to update the FW version checks?

Yes, will need to change the implementation. This is only the firware change. I will make a PR into this PR before we merge to developer. I am going to have a discussion with Martin now on the implementation details.

thattil:
What I will implement is that the software will first read the fill level (for ADCs, the current ADC muxed in) and read that many samples (until we reach  the corresponding number of samples).  If too few samples, it will be stuck until end of acquisition, after which it will send out the remaining samples.

Question:
Would you like a "asamplesl", "dsamplesl" or "tsamplesl" to read how many samples "left" to make a frame? [eg. dsamplesl = dsamples - current level at dbit fifo level].  This might be interesting to know why you might be stuck (because your pattern had too few samples).

erikfrojdh:
Let's leave out the commands for now and add them only if people need them. Maybe we should document the registers?

mmarti04:
@thattil  @AliceMazzoleni99 
Can we have an enable checkbox/button in the ctb gui for the cont. readout ?
and while i'm already asking for extra buttons ... a checkbox to disable cleanFifos(); in line 2215 of slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c would also be nice. cleanFifos() was broken in the firmware, now it's working again and i don't like it. I can break it in firmware again but would prefer to have an option for this in the software.

mmarti04:
inluded in PR#1303
Will likely not need individual commands but only something like a 32-bit "readout setting" register

PR #1164, opened 2025-03-19
Dev/scikitbuild

erikfrojdh:
- Modified python/CMakesList.txt to work with scikit-build-core
- added pyproject.toml on top level
- removed setup.py (only build the python extension through cmake now )
- split the conda recipes into python-client and main-library
- Messy git history but will be squashed on merge


JamesOHeaDLS:
Hi @erikfrojdh 

Thanks for this - seems to work nicely - I have been able to run a build and pip install the generated wheel file and import slsDet in a Python file.

I have a couple of suggested tweaks to the build files:

* Because scikit-build-core is being used, that can be referenced for the section header in pyproject.toml instead of scikit-build
* In the same file, build.verbose is recommended instead of cmake.verbose
* In CMakeLists.txt, around line 64, there is a note about not being able to use FetchContent_MakeAvailable - I think that it is possible if the following is used to exclude zmq from install

```
# Set the ZMQ_INSTALL option to OFF
set(ZMQ_INSTALL OFF CACHE BOOL "Disable installation of libzmq")

FetchContent_MakeAvailable(libzmq)
```
The only other change I made was to get around an error of the build process not being able to see .clang-tidy, even though I can see the file in the project structure.

```
CMake Error: File /tmp/build-via-sdist-pvjs_r7a/slsdet-2025.3.19/.clang-tidy does not exist.
CMake Error at CMakeLists.txt:331 (configure_file):
  configure_file Problem configuring file


-- Configuring incomplete, errors occurred!

*** CMake configuration failed
```

My workaround was 

```
if(EXISTS ".clang-tidy")
    configure_file(.clang-tidy  
    ${CMAKE_BINARY_DIR}/.clang-tidy  
    )
endif()
```

but I feel there is a path issue somewhere.   Is the file expected to end up inside the slsdet-2025.3.19 area?

Finally, I saw the following warning pop-up a couple of times during the build process

```
CMake Deprecation Warning at /tmp/tmpkxgy6uui/build/_deps/libzmq-src/CMakeLists.txt:7 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```

The build process appears to confuse the version number within the VERSION text file with the version number of CMake, but I couldn't see why?  Maybe some reserved variable name is being used?

Thanks,

James

erikfrojdh:
Hi @JamesOHeaDLS 

Thank you for you feedback. I tried to address your comments as best as I could. 

> - Because scikit-build-core is being used, that can be referenced for the section header in pyproject.toml instead of scikit-build
> - In the same file, build.verbose is recommended instead of cmake.verbose

Thanks, fixed it. 

> In CMakeLists.txt, around line 64, there is a note about not being able to use FetchContent_MakeAvailable - I think that it is possible if the following is used to exclude zmq from install

Unfortunately, this still installs the cmake files for libzmq. I didn't find any other workaround than passing 
EXCLUDE_FROM_ALL to FetchContent_Populate. Good news is that from cmake 3.28 we can pass it to FetchContent_Declare instead. 
 

> but I feel there is a path issue somewhere. Is the file expected to end up inside the slsdet-2025.3.19 area?

I took your workaround. I don't know why the file gets excluded, but it's not important for the build. It's being copied to the build dir to make it easier to run clang-tidy from the command line. 

> Finally, I saw the following warning pop-up a couple of times during the build process 

This comes from libzmq where they specify a very old cmake version. I made a patch to change it to 3.15

erikfrojdh:
Correction on my previous answer scikit-build-core still uses the key scikit-build in pyproject.toml

conda-recipes/main-library/copy_lib.sh
```diff
@@ -4,7 +4,6 @@
 mkdir -p $PREFIX/lib
 mkdir -p $PREFIX/bin
 mkdir -p $PREFIX/include/sls
```
thattil, line 6: Don't we also need to add the following? We must have missed it in the frame synchonizer PR.
cp build/install/bin/slsFrameSynchronizer $PREFIX/bin/.

thattil, line 6: We should add the following as well.
cp build/install/bin/sls_detector $PREFIX/bin/.

thattil (commented):
LGTM. Very minor changes. I tested all our cmake options just in case. Also tested building from source, installing, conda build and pip install. They all give the right versioning.

PR #1167, opened 2025-03-20
Dev/multirxr proper cleanup on ctrl + c and versioning

### LINKED ISSUE #1160: slsMultiReceiver does not terminate Arping processes on exit
##### *Detector type: 
Jungfrau

8.0.2

##### *Describe the bug
After an acquisition is started the Arping process is not terminated on exit

### LINKED ISSUE #1166: Versioning for slsMultiReceiver and slsFrameSynchronizer

thattil:
Since its breaking API, major changes in command line. 
- also added -p or --port (rx_tcpport or -t deprecated) for slsReceiver and slsMultiREceiver
- refactored to remove separate command line options code
- using sls_use_tests to test port 65535 to throw in Receiver constructor to test multiple receivers where one crashes
- multi process: with child processes. if a child throws,it will dstroy semaphore and exit with failure. The parent process will be informed via waitpid and check its exit status and will send sigint to all other child processes so they exit and not continuw with one less child. The parent process itself ignores sigint and waits for all child process to be done to exit gracefully. The child processes get a copy of signal handler so rach chlid calls its own signal handler posting its own copy of semaphore.
- slsFramesync: with threads. parent threads creates vector of semaphores (changing size in  child threads are not thread safe) and each thread is allotted one. Only main thread calls signal handler when ctlr +c.which will post all the sempahores in vector. If a child thread throws, parent thread cannot catch, so child thread catches, destroys semaphore but captures the exception. The parent then waits for all to join,  cleans up gracefully, then checks the captured expression to print and exit with failure or success.
-added support for deprecated options or arguments
- testing the parsed options


erikfrojdh (approved):

slsReceiverSoftware/src/FrameSynchronizerApp.cpp
```diff
@@ -504,51 +493,115 @@ void GetDataCallback(slsDetectorDefs::sls_receiver_header &header,
...
-                        printHeadersLevel = sls::logINFOBLUE;
-                    }
-                }
-            } else
-                throw std::runtime_error("Invalid number of arguments");
-        } catch (const std::exception &e) {
-            cprintf(RED, "Error: %s\n%s\n", e.what(), getHelpMessage().c_str());
-            return EXIT_FAILURE;
+            } catch (...) {
+                throw sls::RuntimeError("Invalid number of receivers." +
```
AliceMazzoleni99, line 539: Maybe one can add in the error message that num receiver can not be larger than 100

slsReceiverSoftware/src/MultiReceiverApp.cpp
```diff
@@ -151,56 +134,173 @@ void GetData(slsDetectorDefs::sls_receiver_header &header,
...
         try {
             if (argc == 3 || argc == 4) {
-                startTCPPort = sls::StringTo<uint16_t>(argv[1]);
-                if (startTCPPort == 0) {
-                    throw std::runtime_error("Invalid start tcp port");
-                }
-                numReceivers = std::stoi(argv[2]);
+                startPort = sls::StringTo<uint16_t>(argv[1]);
+                numReceivers = sls::StringTo<uint16_t>(argv[2]);
                 if (numReceivers > 1024) {
```
AliceMazzoleni99, line 153: In FrameSynchronizerApp the limit for numreceivers is 100

slsReceiverSoftware/src/MultiReceiverApp.cpp
```diff
@@ -233,48 +334,54 @@ int main(int argc, char *argv[]) {
...
+                        << "Registering AcquisitionFinished()";
+                    receiver.registerCallBackAcquisitionFinished(
+                        AcquisitionFinished, nullptr);
+
+                    /* 	- Call back for raw data */
+                    LOG(sls::logINFOBLUE) << "Registering GetData()";
+                    receiver.registerCallBackRawDataReady(GetData, nullptr);
+                }
+
+                /**	- as long as no Ctrl+C */
```
AliceMazzoleni99, line 372: Maybe i dont quite understand. But will this not be a potential deadlock if a thread throws an error and never reaches sem_wait?

slsReceiverSoftware/src/ReceiverApp.cpp
```diff
@@ -22,9 +25,95 @@ void sigInterruptHandler(int p) { sem_post(&semaphore); }
...
+                std::ostringstream oss;
+                oss << "Could not set Effective UID to " << userid << ". Got "
+                    << geteuid();
+                throw sls::RuntimeError(oss.str());
+            }
+            LOG(sls::logINFO) << "Process Effective UID changed to " << userid;
+        }
+    }
 
     // Catch signal SIGINT to close files and call destructors properly
```
AliceMazzoleni99, line 118: Generally loads of duplicate code maybe have a function ProcessCommandLineArguments that can be shared between ReceiverApp, FrameSynchronizerApp, and MultiReceiverApp

slsReceiverSoftware/src/MultiReceiverApp.cpp
thattil: Short answer. 
It wont reach a deadlock, but there are semaphore resource leaks.

Long answer: 
It will not be a deadlock because each child process (for MultiReceiver) will get a copy of the semaphore. Since it never reaches sem_wait and exits with an exception, the parent will wait for child process to change state (waitpid) and so will reap the zombie child. Ctrl+c on the other hand, will cause each child process to call the signal handler which posts the semaphore and hence continues with the execution to clean up. 
However, semaphore resources leaks if it throws before sem_wait.

For FrameSynchronizer, they are using threads (since we need to sync and its easier with threads). Here, main thread creates a vector of semaphores, where each thread has access to its own. But here, need to make this handle an exception. Here, with ctrl+c, only the main thread gets the signal handler to post.
Also, need to handle semphore resource leak.

slsReceiverSoftware/src/ReceiverApp.cpp
thattil: Done

slsReceiverSoftware/src/MultiReceiverApp.cpp
thattil: Done

slsReceiverSoftware/include/sls/Receiver.h
```diff
@@ -13,22 +13,12 @@ class Receiver : private virtual slsDetectorDefs {
...
      * Constructor
-     * Starts up a Receiver server. Reads configuration file, options, and
-     * assembles a Receiver using TCP and UDP detector interfaces
+     * Starts up a Receiver server.
+     * Assembles a Receiver using TCP and UDP detector interfaces
      * throws an exception in case of failure
-     * @param argc from command line
-     * @param argv from command line
+     * @param port TCP/IP port number
      */
```
AliceMazzoleni99, line 20: should this be TCP/UDP port number

thattil: Its TCP port number. The receiver has to start with that so the client can communicate with it. 

The UDP port number just needs to be in the config file as the client can then configure the receiver and detector module however it is in the config file.

slsReceiverSoftware/src/CommandLineOptions.cpp
```diff
@@ -0,0 +1,357 @@
...
+            handleCommonOption(opt, optarg, base);
+            break;
+        case 'c':
+        case 'n':
+        case 't':
+            handleAppSpecificOption(opt, optarg, base, multi, frame);
+            break;
+        default:
+            throw sls::RuntimeError("Invalidddd arguments." + getHelpMessage());
+        }
```
AliceMazzoleni99, line 55: minor Typo Invalid

slsReceiverSoftware/src/CommandLineOptions.cpp
```diff
@@ -0,0 +1,357 @@
...
+}
+
+ParsedOptions CommandLineOptions::parse(int argc, char *argv[]) {
+    CommonOptions base;
+    MultiReceiverOptions multi;
+    FrameSyncOptions frame;
+    base.port = DEFAULT_TCP_RX_PORTNO;
+
+    auto optString = buildOptString();
+    auto longOptions = buildOptionList();
```
AliceMazzoleni99: Have those two in constructor and store as member variable

slsReceiverSoftware/src/CommandLineOptions.cpp
```diff
@@ -0,0 +1,357 @@
...
+
+    LOG(sls::logWARNING)
+        << "Deprecated options will be removed in future versions. "
+           "Please use the new options.\n";
+
+    // default deprecated values
+    if (nargs == 1) {
+        return std::make_tuple(DEFAULT_TCP_RX_PORTNO, 1, false);
+    }
+
```
AliceMazzoleni99, line 262: What is args[0]?

thattil: The name of the binary (slsReceiver, slsMultiReceiver or slsFrameSynchronizer)

slsReceiverSoftware/src/CommandLineOptions.cpp
thattil: Good one. Done.

slsReceiverSoftware/src/CommandLineOptions.cpp
AliceMazzoleni99: mmh so in getopt_long how is the binary treated does it return -1

thattil: it skips it automatically because it knows it will be the binary name as it is by convention.

slsReceiverSoftware/src/CommandLineOptions.cpp
```diff
@@ -0,0 +1,359 @@
...
+        LOG(sls::logINFO) << "Number of receivers: " << frame.numReceivers;
+        LOG(sls::logINFO) << "Print headers: " << frame.printHeaders;
+    }
+
+    switch (appType_) {
+    case AppType::SingleReceiver:
+        return base;
+    case AppType::MultiReceiver:
+        static_cast<CommonOptions &>(multi) = base;
+        return multi;
```
AliceMazzoleni99, line 104: nice reference cast I didn't know one can do that

slsReceiverSoftware/src/CommandLineOptions.cpp
```diff
@@ -0,0 +1,359 @@
...
+                                    "move away from the deprecated options.\n";
+        }
+
+        // unsupported deprecated arguments
+        if (appType_ == AppType::SingleReceiver) {
+            throw sls::RuntimeError("Invalid arguments." + getHelpMessage());
+        }
+
+        // parse deprecated arguments
+        std::vector<std::string> args(argv, argv + argc);
```
AliceMazzoleni99, line 76: why do you make a copy then? - Is this necessary or can we just pass argv directly

thattil: Just to make it easier to test the funciton directly if its a vector of strings. I can just pass {"ddd", "dfgf","ddf"} in testing it

slsReceiverSoftware/src/MultiReceiverApp.cpp
```diff
@@ -146,164 +124,115 @@ void GetData(slsDetectorDefs::sls_receiver_header &header,
...
+                // each child process gets a copy of the semaphore
+                sem_wait(&semaphore);
+                sem_destroy(&semaphore);
+
             } catch (...) {
+                sem_destroy(&semaphore);
```
AliceMazzoleni99, line 220: Here https://pubs.opengroup.org/onlinepubs/009695399/functions/sem_destroy.html it says that if processes are still waiting on a semaphore sem_destroy is undefined. I would let the parent process destroy it

slsReceiverSoftware/src/MultiReceiverApp.cpp
```diff
@@ -146,164 +124,115 @@ void GetData(slsDetectorDefs::sls_receiver_header &header,
...
+
+                    /* 	- Call back for raw data */
+                    LOG(sls::logINFOBLUE) << "Registering GetData()";
+                    receiver.registerCallBackRawDataReady(GetData, nullptr);
+                }
+
+                /**	- as long as no Ctrl+C */
+                // each child process gets a copy of the semaphore
+                sem_wait(&semaphore);
+                sem_destroy(&semaphore);
```
AliceMazzoleni99, line 215: also here I would the parent process destroy it

slsReceiverSoftware/src/CommandLineOptions.cpp
```diff
@@ -0,0 +1,359 @@
...
+               "Disabled by default.\n" +
+               "\t-u, --uid           : Set effective user id if receiver "
+               "started "
+               "with privileges. \n\n";
+    }
+    throw sls::RuntimeError("Unknown AppType for help message");
+}
+
+void CommandLineOptions::setupSignalHandler(int signal, void (*handler)(int)) {
+    // Catch signal SIGINT to close files and call destructors properly
```
AliceMazzoleni99: maybe this function should be in a common or utils file and not a class member function

thattil: Yea i too felt its a bit off here. will move it.

slsReceiverSoftware/src/MultiReceiverApp.cpp
thattil: As far as i understand, the parent process has no access to it anymroe since each child got its own copy.. i think i might have to use sigint from the parent process somehow. will try something.

slsReceiverSoftware/src/Receiver.cpp
```diff
@@ -29,6 +29,9 @@ Receiver::~Receiver() = default;
 
 Receiver::Receiver(uint16_t port) {
     validatePortNumber(port);
+    if (port == 9999) {
+        throw sls::RuntimeError("Throwing for testing purposes. ");
+    }
```
AliceMazzoleni99, line 35: minor wrap in USE_TESTS

thattil: Done

slsReceiverSoftware/src/CommandLineOptions.cpp
thattil: Moved to network_utils

slsReceiverSoftware/src/MultiReceiverApp.cpp
thattil: parent cannot have access to the child process semaphore. So solution now is, the child deletes its own when it throws or when it exits normally and when the child process exits with a failure, this is picked up by the parent when using "waitpid" and then it iwll send a sigint to all the other child processes, so they dont continue with one child missing.

slsReceiverSoftware/src/MultiReceiverApp.cpp
thattil: they wont be wiating on a semaphore as they would have crashed before reaching the sem_wait

slsReceiverSoftware/src/CommandLineOptions.cpp
thattil: me neither :)

AliceMazzoleni99 (approved):

PR #1193, opened 2025-04-22
Dev/automate version number

AliceMazzoleni99:
Automates Version number in the conda metadata and toml file. Note that one has to run the script before building the conda packages (it updates the version in the pyproject,toml file and the version number in the VERSION file).  
`python update_version.py version_number`
version number needs to be in the format year.month.day


thattil (changes requested):
Looks very good. Just one minor thing. with conda 10.0.2-dev1 doesnt work. it doesnt seem to like the hypen '-'. 10.0.2.dev1 worked. For normal build and instlalation, both worked well with '-' or '.'. and release in 'versions' command rightly gave the whole thing. cmake compiled with 10.0.2. 

so the only fix is just conda to work with '-' in versioning.

AliceMazzoleni99:
Oke conda build only takes version numbers which follow the PEP 440 specifications (which do not include - characters). Ive updatet the update_version script to normalize the passed version to the PEP 440 standard. https://peps.python.org/pep-0440/

pyproject.toml
```diff
@@ -1,27 +1,27 @@
 [build-system]
-requires = ["scikit-build-core>=0.10", "pybind11", "numpy"]
+requires = [ "scikit-build-core>=0.10", "pybind11", "numpy",]
```
erikfrojdh, line 2: Note for later, we need to set a common formatting to avoid these diffs...

pyproject.toml
```diff
@@ -1,27 +1,27 @@
...
 build-backend = "scikit_build_core.build"
 
 [project]
 name = "slsdet"
-version = "2025.3.19"
-
+version = "0.0.0"
```
erikfrojdh, line 7: scikit-build-core: also supports dynamic versioning like in the conda meta.yaml. I suggest using a regex here as well: https://scikit-build-core.readthedocs.io/en/latest/configuration/dynamic.html#regex then we don't need the additional script to update version

pyproject.toml
```diff
@@ -1,27 +1,27 @@
...
 
 [tool.cibuildwheel]
 before-all = "uname -a"
 
-[tool.scikit-build]
-build.verbose = true
-cmake.build-type = "Release"
-install.components = ["python"]
+[tool.scikit-build.build]
```
erikfrojdh, line 12: Any particular reason to split into distinct sections?

update_version.py
```diff
@@ -6,6 +6,8 @@
 
 import sys
 import re
+import toml
```
erikfrojdh, line 9: This file (update_version.py) might not be needed if using regex also in pyproject.toml

erikfrojdh (changes requested):
Nice work using the regex in the meta.yaml. Let's try the same thing in pyproject.toml

pyproject.toml
AliceMazzoleni99: Thats very nice. I changed it however I would still keep the update_version.py file as it standardizes the passed version to PEP 440.

conda-recipes/main-library/meta.yaml
```diff
@@ -1,11 +1,11 @@
...
-  version: 2025.3.19
-
-
 source:
   path: ../..
 
+{% set version = load_file_regex(load_file = 'VERSION', regex_pattern = '(\d+(?:\.\d+)*(?:[\+\w\.]+))').group(1) %}
+package:
+  name: sls_detector_software
+  version: {{ version }} #2025.3.19
```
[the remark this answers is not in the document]
AliceMazzoleni99: removed

conda-recipes/python-client/meta.yaml
```diff
@@ -1,10 +1,11 @@
...
-  name: slsdet
-  version: 2025.3.19 #TODO! how to not duplicate this?
-
 source:
   path: ../..
 
+{% set version = load_file_regex(load_file = 'VERSION', regex_pattern = '(\d+(?:\.\d+)*(?:[\+\w\.]+))').group(1) %}
+package:
+  name: slsdet
+  version: {{ version }} #
```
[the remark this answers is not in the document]
AliceMazzoleni99: removed

pyproject.toml
AliceMazzoleni99: Cant remember to have changed that to be honest.

erikfrojdh: Maybe your editor playing games?

erikfrojdh (approved):

thattil:
> Just my comments for testing later:
> 
> Ways that the version is exposed
> 
> * slsdet.**version**
> * importlib.metadata.version('slsdet')
> * sls_detector_get --version #reports developer + date
> 
> Code that uses the version:
> 
> * CMakeLists.txt
> * cmake/project_version.cmake
> * conda-recipe/meta.yaml
> * pyproject.toml

sls_detector_get --version(slsReceiver, other sls_detector* binaries) all require to run the updateClientAPIVersion.sh which updates the versionAPI.h to have 'developer' if empty or the semantic versioning with date. I have made a new issue 'Part 2 of automatic versioning: client and receiver binary versioning #1196' #1196 which will address this with a python script instead and by changing the 'deve'oper' to '0.0.0' for example and this will have to be fixed accordingly in the code as well for the version comparisons.
 
Similary, for the server versions.

thattil:
> > Just my comments for testing later:
> > Ways that the version is exposed
> > 
> > * slsdet.**version**
> > * importlib.metadata.version('slsdet')
> > * sls_detector_get --version #reports developer + date
> > 
> > Code that uses the version:
> > 
> > * CMakeLists.txt
> > * cmake/project_version.cmake
> > * conda-recipe/meta.yaml
> > * pyproject.toml
> 
However, one thing one can test is the package version from "versions" command or the API getPackageVersion. This should be updated when using "update_version.py".

thattil (approved):

PR #1199, opened 2025-04-29
Dev/test all acquire file

thattil:


slsDetectorSoftware/tests/Caller/test-Caller-eiger.cpp
```diff
@@ -17,6 +17,62 @@ namespace sls {
...
+        testCommonDetAcquireInfo det_config;
+        det_config.num_frames_to_acquire = num_frames_to_acquire;
+        set_common_acquire_config_state(det, det_config);
+
+        // set default specific det type config
+        det.setReadNRows(256);
+        det.setDynamicRange(16);
+
+        // acquire
+        test_acquire_with_receiver(caller, std::chrono::seconds{2});
```
AliceMazzoleni99, line 54: just to clarify: how does one choose timeout

slsDetectorSoftware/tests/Caller/test-Caller-global.cpp
```diff
@@ -88,4 +88,91 @@ void test_onchip_dac_caller(defs::dacIndex index, const std::string &dacname,
...
+}
+
+void test_frames_caught(const Detector &det, int num_frames_to_acquire) {
+    auto frames_caught = det.getFramesCaught().tsquash(
+        "Inconsistent number of frames caught")[0];
+    REQUIRE(frames_caught == num_frames_to_acquire);
+}
+
+void test_acquire_with_receiver(Caller &caller, std::chrono::seconds timeout) {
+    REQUIRE_NOTHROW(caller.call("rx_start", {}, -1, PUT));
```
AliceMazzoleni99, line 136: reciver_id in Caller::call has no default argument why does one not need to pass it.

slsDetectorSoftware/tests/Caller/test-Caller-global.cpp
```diff
@@ -88,4 +88,91 @@ void test_onchip_dac_caller(defs::dacIndex index, const std::string &dacname,
...
+    REQUIRE_NOTHROW(caller.call("start", {}, -1, PUT));
+    std::this_thread::sleep_for(timeout);
+    REQUIRE_NOTHROW(caller.call("rx_stop", {}, -1, PUT));
```
AliceMazzoleni99, line 139: Just for clarification why wouldnt one stop the detector and thus the acquisition process

thattil: because the detector will stop on its own as we have only put 2ms as period and 2 frames. it should be done in the timeout we have given (2 seconds).

slsDetectorSoftware/tests/Caller/test-Caller-global.cpp
thattil: receiver_id is only used for udp_dstlist command where we give many destinations to put into a round robin list.

slsDetectorSoftware/tests/Caller/test-Caller-eiger.cpp
thattil: this could have been also 1 second.. it should be less than number of frames * period.  Normallly, you can use get status to be sure or acquire. This would be the right way of doing it, but we are not testing that.

slsDetectorSoftware/tests/Caller/test-Caller-global.cpp
```diff
@@ -88,4 +88,91 @@ void test_onchip_dac_caller(defs::dacIndex index, const std::string &dacname,
...
+                                   uint64_t num_frames_to_acquire,
+                                   uint64_t expected_image_size) {
+    assert(file_info.file_format == defs::BINARY);
+    std::string fname = file_info.file_path + "/" + file_info.file_prefix +
+                        "_d0_f0_" + std::to_string(file_info.file_acq_index) +
+                        ".raw";
+    uint64_t expected_file_size =
+        num_frames_to_acquire *
+        (expected_image_size + sizeof(defs::sls_receiver_header));
+    std::cout << "exepected file size: " << expected_file_size
```
AliceMazzoleni99, line 122: maybe replace std::cout with a log

thattil: ah was supposed to remove thiis line actually. Thanks!

slsDetectorSoftware/tests/Caller/test-Caller-gotthard2.cpp
```diff
@@ -17,6 +17,63 @@ namespace sls {
...
+
+        // acquire
+        test_acquire_with_receiver(caller, std::chrono::seconds{2});
+
+        // check frames caught
+        test_frames_caught(det, num_frames_to_acquire);
+
+        // check file size (assuming local pc)
+        // num_channels * num_chips * bytes_per_pixel * num_frames
+        size_t expected_image_size = 128 * 10 * 2;
```
AliceMazzoleni99, line 62: I think we should store those numbers like num_chips and num_channels, bytes_per_pixel in a gotthard_defs file or something

AliceMazzoleni99: same for all the other detectors

slsDetectorSoftware/tests/Caller/test-Caller-jungfrau.cpp
```diff
@@ -15,6 +15,63 @@ namespace sls {
...
+        // check frames caught
+        test_frames_caught(det, num_frames_to_acquire);
+
+        // check file size (assuming local pc)
+        size_t expected_image_size = 0;
+        // pixels_row_per_chip * pixels_col_per_chip * num_chips *
+        // bytes_per_pixel
+        if (num_udp_interfaces == 1) {
+            expected_image_size = 256 * 256 * 8 * 2;
+        } else {
```
AliceMazzoleni99, line 60: Is 256 the num of rows or 512?

slsDetectorSoftware/tests/Caller/test-Caller-mythen3.cpp
```diff
@@ -17,6 +17,61 @@ namespace sls {
...
+        testFileInfo test_file_info;
+        set_file_state(det, test_file_info);
+        testCommonDetAcquireInfo det_config;
+        det_config.num_frames_to_acquire = num_frames_to_acquire;
+        set_common_acquire_config_state(det, det_config);
+
+        // set default specific det type config
+        det.setDynamicRange(16);
+        int test_counter_mask = 0x1;
+        int num_counters = __builtin_popcount(counter_mask);
```
AliceMazzoleni99, line 50: Should this be test_counter_mask

slsDetectorSoftware/tests/Caller/test-Caller-chiptestboard.cpp
```diff
@@ -17,6 +17,257 @@ namespace sls {
 using test::GET;
 using test::PUT;
 
+struct testCtbAcquireInfo {
+    defs::readoutMode readout_mode{defs::ANALOG_AND_DIGITAL};
+    bool teng_giga{false};
```
AliceMazzoleni99, line 22: typo: ten_giga

slsDetectorSoftware/tests/Caller/test-Caller-global.cpp
thattil: removed the print out. :+1:

slsDetectorSoftware/tests/Caller/test-Caller-jungfrau.cpp
thattil: its as in the comment. its rows/col per chip. its 256 for both. but when i use slsDetectorDefs::detparameters as you suggested in a previous comment, it will be more clear.

slsDetectorSoftware/tests/Caller/test-Caller-chiptestboard.cpp
```diff
@@ -17,6 +17,257 @@ namespace sls {
...
+    test_frames_caught(det, num_frames_to_acquire);
+
+    // calculate image size
+    uint64_t num_analog_bytes = 0, num_digital_bytes = 0,
+             num_transceiver_bytes = 0;
+    if (test_info.readout_mode == defs::ANALOG_ONLY ||
+        test_info.readout_mode == defs::ANALOG_AND_DIGITAL) {
+        uint32_t adc_enable_mask =
+            (test_info.teng_giga ? test_info.adc_enable_1g
+                                 : test_info.adc_enable_10g);
```
AliceMazzoleni99: Shouldnt it be if test_info.teng_giga == True then adc_enable_mask = test_info.adc_enable_10g

slsDetectorSoftware/tests/Caller/test-Caller-mythen3.cpp
thattil: good catch. thanks!

slsDetectorSoftware/tests/Caller/test-Caller-gotthard2.cpp
thattil: will pick them up from slsDetectorDefs::detParameters. bytes per pixel can be calculated from dynamic range. Will change this.

slsDetectorServers/xilinx_ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1577,10 +1582,12 @@ void *start_timer(void *arg) {
...
             // fill data
+            int remaining = imageSize - dataSent;
+            int dataSize = remaining < maxDataSize ? remaining : maxDataSize;
             memcpy(packetData + sizeof(sls_detector_header),
-                   imageData + srcOffset,
-                   (imageSize < maxDataSize ? imageSize : maxDataSize));
-            srcOffset += maxDataSize;
+                   imageData + srcOffset, dataSize);
+            srcOffset += dataSize;
+            dataSent += dataSize;
```
AliceMazzoleni99, line 1591: Ive noticed that packetdata is also allocated on the stack

thattil: packetdata is always max 9k because of jumbo frames.

PR #1201, opened 2025-04-30
added check if reciever is running

### LINKED ISSUE #1198: Setting transceiver, analog and digital samples should be blocked when receiver is running
**Current behaviour:** (developer + all branches supporting a/d/t samples)
```bash
p rx_start
p tsamples 200
p start
#segfault
```

Setting the transceiver samples, or anything that changes the data size and forces the receiver to reallocate the frame buffer needs to be blocked if the receiver is started. 

**Expected behaviour:** 
```bash
p rx_start
p tsamples 200
#ERROR: Can not set tsamples when the receiver is active
```

AliceMazzoleni99:
- check if receiver is idle before updating num samples


slsDetectorSoftware/tests/Caller/test-Caller-rx-running.cpp
```diff
@@ -0,0 +1,239 @@
...
+}
+
+TEST_CASE("general - cant put if receiver is not idle", "[.cmdcall][.rx]") {
+
+    Detector det;
+    Caller caller(&det);
+
+    {
+        auto prev_frames =
+            det.getNumberOfFrames().tsquash("#bursts should be same to test");
```
AliceMazzoleni99: typo: #num frames should be same to test

AliceMazzoleni99 (commented):
Looks good. Much better test coverage then before. I'm not sure if we should test for everything e.g. udp_dstip, rx_tcpport.

thattil:
> Looks good. Much better test coverage then before. I'm not sure if we should test for everything e.g. udp_dstip, rx_tcpport.

I only tested the parameters that change the fifo structure as that was the original issue.

PR #1207, opened 2025-05-08
ctb: fix bug in blackfin read access to firmware registers

mmarti04:
There is a rare bug when accessing registers on the firmware if they are not in the CPU's clock domain

bus_w(REG_ADDR, A);
B = bus_r(REG_ADDR);
--> for a few very specific REG_ADDR's and a very low run_clk frequency we can get A!=B

We can do something about that by introducing a waitrequest from the firmware towards the blackfin. The connection for this exists but is currently configured to be ignored. To change that one could recompile the blackfin kernel and manually update all of them or do the thing below.

Maybe at some point also in Jungfrau & Moench if Carlos needs his waitrequest to work at some point. For now just relevant for the CTB.


mmarti04:
found some more problems with this, likely firmware issue. Do not merge for the moment

mmarti04:
can be merged, but register access on blackfin detectors should be quickly tested afterwards. I checked with Carlos and this should not change anything there.

thattil:
@mmarti04 I made some refactoring and updated the binary. Please have a look.

thattil (approved):
@mmarti04 I made some refactoring and updated the binary. Please have a look.

thattil:
As discussed, the other detector servers using blackfin (jungfrau, moench) do not need this at the moment.

PR #1209, opened 2025-05-15
Dev/automate version part2

AliceMazzoleni99:
updates version api according to version in VERSION file


slsSupportLib/src/Version.cpp
```diff
@@ -21,7 +21,8 @@ Version::Version(const std::string &s) {
 }
 
 bool Version::hasSemanticVersioning() const {
-    return version_ != defaultBranch_;
+
+    return (version_ != defaultBranch_[0]) && (version_ != defaultBranch_[1]);
```
thattil, line 25: would you change this to defaultVersion? because defaultBranch does not make much sense anymore..

thattil (changes requested):
Tested it against different versions. It looks good! Just some minor comments.

thattil (approved):

PR #1242, opened 2025-06-16
Roi per port

### LINKED ISSUE #1139: at least single rx_roi for each port
Independent rx_rois per module, please!

##### *Detector type: 
Mostly JF

8.0.2

##### Priority:
Enourmously High 

##### *State the feature:
Independent rx_rois per module

##### Is your feature request related to a problem. Please describe:
Problem is that there is too rarely Raclette at the Oase

thattil:
API:
- breaking API: changed to allow vector of ROIS at multi level to set, to get (individually or multi level), clearroi
- when converting multi roi vectors to port level values (throw for any overlap or more than 1roi per port)
- individual roi is a vector of size 1 or 2 depending on number of interfaces and the channel values are local port values.
- getroi multi level is just picked up from reicever master (has multi level roi metadata for master file)
- added 'getPortSize' and 'getPortPerModuleGeometry' in Detector API
- setting number of interfaces clears roi
- upgraded shm version (single roi is removed from shm)

Internally:
- multi roi vector always has size atleast 1, with -1 or 0s to differentitate "complete roi" or "outside of roi" and other values for "partial roi"
- port roi vector always has size of number of interfaces (1 for each port). Again with -1s,0s or port level values.

Command line:
- rx_roi action cannot be inferred (has to be p or g) 
- compatible with old parsing `xmin xmax ymin ymax` for 1 roi. 
- For multi rois  Use square brackets and space between Rois. Use commas and no space inside an roi=> [xmin,xmax,ymin,ymax] [xmin,xmax,ymin,ymax]. 
- If you use semicolon between rois, use quotes. "[xmin,xmax,ymin,ymax];[xmin,xmax,ymin,ymax]"

receiver master file
- metadata multi roi converted to String as other arrays.
- binary master version 7.3 ->8.0, hdf5 master version: 6.7 ->7.0
- implemented hdf5 mapping for virtual file with roi (dataset for each roi)

slsDetectordefs:
- added 'overlap' function in roi structure to reuse

tests:
- rx_roi command tested for syntax, individual roi and also acquired to check for presence of master and virtual file (hdf5)
- python scripts (test_roi.py) with different module configurations (and number of interfaces) calling rx_roi command test

Python:
- recompiled python bindings
- no specific command

Gui:
- gui displays multi level ROI in yellow boxes (picked up from receiver before acquiring) =>detectorData structure changed to remove ROI and remove extra constructor to incorporate roi


python/src/detector.cpp
```diff
@@ -924,15 +931,17 @@ void init_det(py::module &m) {
...
-                       (Result<defs::ROI>(Detector::*)(sls::Positions) const) &
-                           Detector::getIndividualRxROIs,
-                       py::arg());
     CppDetectorApi.def("getRxROI",
-                       (defs::ROI(Detector::*)() const) & Detector::getRxROI);
-    CppDetectorApi.def(
-        "setRxROI", (void (Detector::*)(const defs::ROI)) & Detector::setRxROI,
-        py::arg());
+                       (std::vector<defs::ROI>(Detector::*)() const) &
+                           Detector::getRxROI);
```
AliceMazzoleni99: Is there a cpp function taking no arguments

slsDetectorGui/include/qDrawPlot.h
```diff
@@ -176,6 +177,7 @@ class qDrawPlot : public QWidget, private Ui::PlotObject {
     uint32_t gainMask{0};
     int gainOffset{0};
     bool gotthard25;
+    std::vector<slsDetectorDefs::ROI> roi{1};
 };
```
AliceMazzoleni99, line 181: should this be size or initializer list - why initialized

python/src/detector.cpp
thattil: Nice catch! had refactored and this function does not exist anymore. Will run the pybind script again and remove this function.

slsDetectorGui/include/qDrawPlot.h
thattil: it default initializes with size, but I can see how this can be confusing. I will remove the size and create the default initialized size in "SetDetectorType".

slsDetectorGui/src/qDrawPlot.cpp
```diff
@@ -971,30 +981,27 @@ void qDrawPlot::Update1dPlot() {
...
         } else {
-            plot1d->EnableRoiBox(std::array<int, 4>{
-                rxRoi.xmin, rxRoi.xmax, (int)plot1d->GetYMinimum(),
-                (int)plot1d->GetYMaximum()});
+            plot1d->EnableRoiBoxes(roi, (int)plot1d->GetYMinimum(),
+                                   (int)plot1d->GetYMaximum());
             if (isGainDataExtracted) {
-                gainplot1d->EnableRoiBox(
-                    std::array<int, 4>{rxRoi.xmin, rxRoi.xmax, 0, 3});
+                gainplot1d->EnableRoiBoxes(roi, 0, 3);
```
AliceMazzoleni99, line 996: Maybe better to wrap this in an if else

slsDetectorGui/src/qDrawPlot.cpp
```diff
@@ -971,30 +981,27 @@ void qDrawPlot::Update1dPlot() {
...
         }
     }
     // ymin and ymax could change (so replot roi every time)
-    if (!rxRoi.completeRoi()) {
-        plot1d->EnableRoiBox(std::array<int, 4>{rxRoi.xmin, rxRoi.xmax,
-                                                (int)plot1d->GetYMinimum(),
-                                                (int)plot1d->GetYMaximum()});
+    if (hasRoi) {
+        plot1d->EnableRoiBoxes(roi, (int)plot1d->GetYMinimum(),
+                               (int)plot1d->GetYMaximum());
```
AliceMazzoleni99, line 1004: wont we care here about GainDataextracted

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -719,49 +719,131 @@ std::string Caller::rx_zmqip(int action) {
...
+        "\tInside each bracket, no spaces allowed.\n\n"
+
+        "\tIf you use semicolon (along with '['and ']' to separate rois), \n"
+        "\tenclose the entire list in quotes.\n"
+        "\tExamples:\n"
+        "\t  [0,100,0,100] [200,300,0,100]\n"
+        "\t  \"[0,100,0,100];[200,300,0,100]\"\n\n"
+
+        "\tNotes:\n"
+        "\t- ROIs can only be set at the multi-module level.\n"
```
AliceMazzoleni99, line 741: Minor: maybe add multiple rois

python/src/detector.cpp
thattil: done

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: updated to include about 1 ROI per udp port and overlapping.

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1696,202 +1666,214 @@ void DetectorImpl::verifyUniqueHost(
...
+}
+
+defs::xy DetectorImpl::getPortGeometry() const {
+    defs::xy portGeometry(1, 1);
+    switch (shm()->detType) {
+    case EIGER:
+        portGeometry.x = modules[0]->getNumberofUDPInterfacesFromShm();
+        break;
+    case JUNGFRAU:
+    case MOENCH:
```
AliceMazzoleni99, line 1744: doesnt jungfrau also support two udp interfaces? Or will it then be two different modules

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1696,202 +1666,214 @@ void DetectorImpl::verifyUniqueHost(
...
+        break;
+    }
+    return portGeometry;
+}
+
+defs::xy DetectorImpl::calculatePosition(int moduleIndex) const {
+    int maxYMods = shm()->numberOfModules.y;
+    int y = (moduleIndex % maxYMods);
+    int x = (moduleIndex / maxYMods);
+    return defs::xy{x, y};
```
AliceMazzoleni99, line 1758: I always thought its row major

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1696,202 +1666,214 @@ void DetectorImpl::verifyUniqueHost(
...
-                defs::xy pos = calculatePosition(iModule, geometry);
-                defs::ROI moduleFullRoi{};
-                moduleFullRoi.xmin = numChansPerMod.x * pos.x;
-                moduleFullRoi.xmax = numChansPerMod.x * (pos.x + 1) - 1;
-                if (is2D) {
-                    moduleFullRoi.ymin = numChansPerMod.y * pos.y;
-                    moduleFullRoi.ymax = numChansPerMod.y * (pos.y + 1) - 1;
-                }
+    for (size_t iModule = 0; iModule < modules.size(); ++iModule) {
+        auto moduleGlobalRoi = getModuleROI(iModule);
```
AliceMazzoleni99, line 1853: Minor: Maybe better to default initialize the portrois iterate over rois and update the portRois instead of iterating over iModule

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -719,49 +719,131 @@ std::string Caller::rx_zmqip(int action) {
...
+        "\t- ROIs coordinates assume no gap pixels, even if they are enabled "
+        "in gui.\n"
+        "\t- To retrieve ROIs per port, specify the module ID when using the "
+        "get command.\n"
+        "\t- Use the command 'rx_clearroi' to clear all ROIs.\n"
+        "\t- Changing the number of UDP interfaces will automatically clear "
+        "the current ROIs.\n\n"
```
AliceMazzoleni99, line 750: Will det->clearRxRoi be called?

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: Yes, 2 ports

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: its column major

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: Changing number of interfaces will change them at the receiver side and also clear the rois in the receiver side. The rois are not stored in the client shared memory or the detector. det->clearRxRoi is only called when calling it directly from the Detector API or from the rx_clearroi command.

slsReceiverSoftware/src/Implementation.cpp
```diff
@@ -155,6 +155,11 @@ void Implementation::setDetectorType(const detectorType d) {
         break;
     }
 
+    // number of portrois should be equal to number of interfaces
+    if (d == EIGER) {
+        ResetRois();
+    }
```
AliceMazzoleni99: why is this explicitly set for Eiger but not for example Jungfrau which might also have two ports?

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: I would not do this at this moment. It can be refactored at a later point as time is of the essence for the next release.

slsReceiverSoftware/src/Implementation.cpp
thattil: You should refresh this as it is not the case anymore. I had modified this yesterday when we came across the portRois declaration with size. Earlier it was done especially for Eiger because it was the only that starts up with 2 udp interfaces by default. Since yesterday, it has been modified to set the size for all detector types as it starts out with size 0.

slsDetectorSoftware/tests/Caller/test-Caller-rx.cpp
```diff
@@ -477,54 +477,309 @@ TEST_CASE("rx_roi", "[.cmdcall]") {
...
+            det.setFileNamePrefix("test");
+
+            det.setAcquisitionIndex(0);
+            det.setFileFormat(defs::BINARY);
+            REQUIRE_NOTHROW(caller.call("acquire", {}, -1, PUT));
+            std::string file_path = "/tmp/test_master_0.json";
+            REQUIRE(std::filesystem::exists(file_path) == true);
+
+            det.setAcquisitionIndex(0);
+            det.setFileFormat(defs::HDF5);
```
AliceMazzoleni99, line 764: Maybe wrap this in #ifdef HDF5C

thattil: Done

AliceMazzoleni99 (approved):

PR #1250, opened 2025-07-29
Dev/doc data format

### LINKED ISSUE #1236: document about data format

thattil:
data format documentation for different detectors


docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+
+.. image:: images/Eiger_module.png
+   :target: _images/Eiger_module.png
+   :width: 350px
+   :align: center
+   :alt: Eiger Module Default Configuration
+
+
+Each Eiger module has two independent identical readout systems (other than firmware), each with its own control port and ``hostname`` to be configured with. They are referred to as the 'top' and 'bottom' half modules. The bottom half module is flipped vertically.
+
```
AliceMazzoleni99, line 144: flipped vertically and horizontally

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+Pixel width
+^^^^^^^^^^^^^
+
+The pixel width can be configured to 4, 8, 16 (default) or 32 bits using the command ``dr``. This affects image size per UDP port or file.
+
+Flip rows
+^^^^^^^^^^
+
+One can use the command ``fliprows`` along with a ``true`` to flip the rows vertically for the bottom or top half module. It is sent out to the reciever, but does not flip rows in the output file itself, but rather streams out this info via the json header and thus instructs the GUI to display them correctly.
+
```
AliceMazzoleni99, line 206: which one bottom or top half?

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+1G/ 10GbE Interfaces
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Each UDP port in the figure above has a 1G and a 10GbE interface. The 1G interface is used for control and configuration, but also for data transmission, while the 10GbE interface is used only for data transmission. The 1GbE interface is enabled by default, but can be enabled using the ``tengiga`` command and updating the appropriate udp_* commands. This setting only affects packetsize and number of packets, but does not affect the total image size. 
+
```
AliceMazzoleni99, line 211: Maybe better: The 1GbE interface is enabled by default. It can be disabled by enabling the ``tengiga`` command and updating the appropritae udp_* commands respectively e.g. udp_dstport2_tengiga (what is the command?). Can one use different bandwidth for different udp ports?

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+Reducing network load
+^^^^^^^^^^^^^^^^^^^^^^
+
+**Activate**: By default, the ``hostname`` command activates the half module, and therefore, all UDP ports are enabled. One can deactivate a whole half module (ie. 2 UDP ports) using command  ``activate`` with the ``false`` argument. This will disable both UDP ports of the half module and no data will be transmitted from it.
+
```
AliceMazzoleni99, line 216: so will i only get half of the acquisition?

AliceMazzoleni99: Maybe clearer By default, the ``hostname``command activates both half modules.

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+
+**Datastream**: The ``datastream`` command with arguments ``left`` or ``right``, along with ``false`` can be used to disable the data streaming from one or all of the two UDP ports in a half module. This allows for more flexible configurations, such as reading only two chips or one UDP port of a half module. 
+
```
AliceMazzoleni99, line 219: typo: both instead of all. How can I specify for which half module i want to set datastream?

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+
+
+Total image size per UDP Port = 8,192 bytes 
+   - 2 chips (1 x 2 grid)
+   - **8** x 256 pixels (chip size: **8 rows**)
+   - 2 bytes (default pixel width)
+
+Note: Still in prototype stage, writes complete image (padded or not depending on ``framepadding`` parameter) to file. Only the summary written to console in the receiver handles the read n rows to calculate complete images received. Only reduces network load, not file size. Use ``rx_roi`` for file size.
+
```
AliceMazzoleni99, line 255: ``readnrows``

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+**Activate**: By default, the ``hostname`` command activates the half module, and therefore, all UDP ports are enabled. One can deactivate a whole half module (ie. 2 UDP ports) using command  ``activate`` with the ``false`` argument. This will disable both UDP ports of the half module and no data will be transmitted from it.
+
+
+**Datastream**: The ``datastream`` command with arguments ``left`` or ``right``, along with ``false`` can be used to disable the data streaming from one or all of the two UDP ports in a half module. This allows for more flexible configurations, such as reading only two chips or one UDP port of a half module. 
+
+Note: Only the activated ports will write data as it does not make sense to write an empty file.
+
+**Read Partial Rows**: The number of image rows per port can be adjusted using the ``readnrows`` command. By default, 256 rows are read, but a smaller value centers the readout vertically (e.g., 8 rows reads 4 above and 4 below the center). Increasing the value symmetrically expands the region toward the top and bottom. Permissible values depend on dynamic range and 10GbE enable.
+
```
AliceMazzoleni99, line 223: I dont know if dynamic range should be specified here. How do they depend on 10GbE enabled

AliceMazzoleni99: What happens when one of the halfmodules is disabled. Will it still center? Or will it throw an error?

docs/src/dataformat.rst
AliceMazzoleni99: >What is with quad? 
It is now dealt with in the quad section.

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+   - 10 chips (1 x 10 grid)
+   - 128 channels
+   - 3 counters
+   - 4 bytes (default pixel width)
+
+Counters
+^^^^^^^^^^^
+
+If all 3 counters are enabeld, the frame size for each channel is multiplied by 3. One can disable one or more of the counters using the ``counters`` command. The frame size will then be reduced accordingly.
+
```
AliceMazzoleni99, line 281: Maybe specify how counters is set e.g. 7 to enable all counters or does one pass a bit string?

AliceMazzoleni99: What would be super nice is to have a link to these commands

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+Pixel width
+^^^^^^^^^^^^^
+
+The pixel width can be configured to 8, 16 or 32 (default) bits using the command ``dr``. 32 bits is actually 24 bits in the chip. This setting does affect image size.
+
+1G/ 10GbE Interfaces
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Each module has a 1G and a 10GbE interface. The 1G interface is used for control and configuration, but also for data transmission, while the 10GbE interface is used only for data transmission. The 10GbE interface is enabled by default, but can be disabled using the ``tengiga`` command and updating the appropriate udp_* commands. This setting only affects packetsize and number of packets, but does not affect the total image size. 
+
```
AliceMazzoleni99, line 299: so per default 1G and 10 GbE are enabled?

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+Default Configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. image:: images/Mythen3_module.png
+   :target: _images/Mythen3_module.png
+   :align: center
+   :alt: Mythen3 Module Default Configuration
+
+Each Mythen3 module is a 1D detector with 10 chips, each with 128 channels. Each channel has 3 counters that are enabled by default.
+
```
AliceMazzoleni99, line 270: Are the 3 counters stored consecutively per channel?

docs/src/dataformat.rst
```diff
@@ -0,0 +1,267 @@
...
+   :alt: Jungfrau Module Two Port Configuration
+
+If both interfaces are enabled using the ``numinterfaces`` command on compatible hardware and firmware, the image splits into top and bottom halves sent over two UDP ports:
+
+   - The top half transmits via the inner interface (``udp_dstport2`` and ``udp_dstip2``).
+
+   - The bottom half uses the outer interface(``udp_dstport`` and ``udp_dstip``).
+
+The number of files per module equals the active UDP ports—two files per module when both interfaces are used.
+
```
AliceMazzoleni99, line 43: Maybe have a link to section File Format

AliceMazzoleni99: Or even add a small example top part of acquisition will be written to filename_d1_f1_0.raw and bottom part ro filename_d0_f1_0.raw

AliceMazzoleni99 (commented):
Super helpful and nice documentation. The images make it very clear. Added some comment where things werent clear for me. One remark: if its not a lot of effort it would be nice to cross reference the commands e.g. udp_dstip, fliprows etc - such that one gets redirected to the documentation.

docs/src/dataformat.rst
thattil: The bottom half is only flipped vertically.

docs/src/dataformat.rst
thattil: > which one bottom or top half?

Depends on the user. We flip the bottom half. Some other users flip the top half. In the end you need to flip one or the other.

thattil: > What is with quad?

Good catch. Will fill it in.

docs/src/dataformat.rst
thattil: Ive updated the documentation, but there is a misunderstanding. Both the UDP ports are used for Eiger. Both of them can be set to use either 1GbE or 10GbE for data.

docs/src/dataformat.rst
thattil: done

docs/src/dataformat.rst
thattil: > Maybe clearer By default, the `hostname`command activates both half modules.

There is only one hostname for a half module. But maybe I can write "By default, each `hostname` command activates the respective half modules that it connects to."

thattil: > so will i only get half of the acquisition?

Yes. I will add this to the documentation for better understanding as well.

thattil: Done

docs/src/dataformat.rst
thattil: Done. Added that info as well if it wasnt clear. To specify which half module, use the module index. For Eiger, each half module gets a module index (every hostname that you can communicate with, you will get a module index)

docs/src/dataformat.rst
thattil: > I dont know if dynamic range should be specified here. How do they depend on 10GbE enabled

For Eiger, it does because  depending on the dynamic range and whether 1GbE or !0GbE is enabled, the packet size and number of packets change. This is what decides whether the readnrows 'value' is allowed or not in the detector server.

thattil: > What happens when one of the halfmodules is disabled. Will it still center? Or will it throw an error?

It should as they are separate indepedent half modules. If one is deactivated and readnrows is set to 8, it should read 4 rows from the center.

docs/src/dataformat.rst
thattil: > Maybe specify how counters is set e.g. 7 to enable all counters or does one pass a bit string?

It depends on where you do it. from the command line, it is just a list of the counters you want enabled. In the Detector API, its the mask. I dont think I should write how to use these commands or API here as this is more about the data format and if I start to write how to use the command, it will just get long. I will remove the 'false' or 'true' arguments to be consistent. I think its better I give a link to the command line instead.

thattil: done

docs/src/dataformat.rst
thattil: No. Either one.  already fixed to be more specific

docs/src/dataformat.rst
thattil: > Maybe have a link to section File Format

I have now linked the section to the top where the numebr of files vs udp port is mentioned.

thattil: > Or even add a small example top part of acquisition will be written to filename_d1_f1_0.raw and bottom part ro filename_d0_f1_0.raw

I think the link is a good idea. number of files and file naming is better of in that section.

docs/src/dataformat.rst
thattil: I am not familiar with that. Is it?

docs/src/dataformat.rst
AliceMazzoleni99: But is it then wrongly displayed in the figure?

thattil: Updated documentation

docs/src/dataformat.rst
thattil: Done

AliceMazzoleni99 (approved):

PR #1252, opened 2025-07-29
dev/python_expose_free

### LINKED ISSUE #1251: free shm is not really exposed in python
related to Shared memory outlives free #1171 

should be able to now do both the following tests:

>>> import slsdet
>>> slsdet.freeSharedMemory()

>>> from slsdet import Detector
>>> d = Detector()
>>> d.free()

thattil:
free shm exposed in python as free function and detector function


AliceMazzoleni99 (approved):
Everything looks good.

PR #1253, opened 2025-07-29
python accessing freed shared memory object

### LINKED ISSUE #1171: Shared memory outlives free
The shared memory remains valid even after a "free". This can cause some interesting effects, for example as Roberto saw with the Ctb after a free and reconfigure the dacnames can differ between two objects (or in his case command line and python). But you could also have a situation where a long running python/C++ process points to a different hostname that you see on the command line. 

**The behaviour is in line with the Linux docs:**

https://man7.org/linux/man-pages/man3/shm_open.3.html

The operation of shm_unlink() is analogous to [unlink(2)](https://man7.org/linux/man-pages/man2/unlink.2.html): it
       removes a shared memory object name, and, once all processes have
       unmapped the object, deallocates and destroys the contents of the
       associated memory region.

**Short python example to reproduce the issue:**

```python
from slsdet import Ctb, dacIndex
from _slsdet import freeSharedMemory #fix this in a separate PR 

d = Ctb()
d.hostname = 'localhost'
d.setDacName(dacIndex(5), 'hej')
freeSharedMemory()

d2 = Ctb()
d2.hostname = 'localhost'
d2.setDacName(dacIndex(6), 'wait')


print(d.daclist)
['DAC0', 'DAC1', 'DAC2', 'DAC3', 'DAC4', 'hej', 'DAC6', 'DAC7', 'DAC8', 'DAC9', 'DAC10', 'DAC11', 'DAC12', 'DAC13', 'DAC14', 'DAC15', 'DAC16', 'DAC17']

print(d2.daclist)
['DAC0', 'DAC1', 'DAC2', 'DAC3', 'DAC4', 'DAC5', 'wait', 'DAC7', 'DAC8', 'DAC9', 'DAC10', 'DAC11', 'DAC12', 'DAC13', 'DAC14', 'DAC15', 'DAC16', 'DAC17']

```

Proposed solution:

- Keep as is and document the behaviour, it's your fault if you mess up...
- Add a key to the shared memory to check if it's valid. Zero out this key on free. (easy to implement)

thattil:
- added a 'isValid' member in shared memory (also updated shm version) with default true, 
- any access to shared memory() checks also for validity. 
- any free will set this to false (by accessing a raw pointers instead of this one) and then map and remove shm. 
- Any access to shm will then check validity in python.
- tests added

To test:
```python
from slsdet import Ctb
d = Ctb()
d.hostname = 'localhsot'
d.exptime = 1

d.free()
d.exptime  # should now throw an error
```

Another example copy pasted from the attached issue (modified):
my_det.config just needs 
```bash
hostname localhost
```

Configure detector:
```bash
p config my_det.config
```

Open a Python terminal and create a detector object (keep the terminal alive)
```python
from slsdet import Detector
d = Detector()
```

Re-configure detector:
```bash
p config my_det.config
```

Now command the object in the python terminal still points to the old shm while command line access gets the new one. 

**Messing up in the same python terminal:**

```python 
from slsdet import Ctb

def load_config(fname):
    d = Ctb()
    d.config = fname

load_config()
d = Ctb()
d.exptime = 1
load_config()
#d still points to old shm, should throw
d.exptime
```


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -40,19 +45,36 @@ void freeSharedMemory(const int detectorIndex, const int moduleIndex) {
     SharedMemory<sharedDetector> detectorShm(detectorIndex, -1);
     if (detectorShm.exists()) {
         detectorShm.openSharedMemory(false);
-        numDetectors = detectorShm()->totalNumberOfModules;
+        // detectorShm() checks for validity
+        if (auto raw = detectorShm.getRawPointer()) {
+            numDetectors = raw->totalNumberOfModules;
+            raw->isValid = false; // mark as invalid
+        }
         detectorShm.removeSharedMemory();
```
AliceMazzoleni99, line 47: Maybe check if you can directly access the shared memory e.g. detectorShm->isValid = false;

slsDetectorSoftware/tests/test-SharedMemory.cpp
```diff
@@ -120,19 +136,19 @@ TEST_CASE("Create several shared memories", "[detector]") {
...
         v[i].createSharedMemory();
-        *v[i]() = i;
-        CHECK(*v[i]() == i);
+        v[i]()->x = i;
+        CHECK(v[i]()->x == i);
     }
 
     for (int i = 0; i != N; ++i) {
-        CHECK(*v[i]() == i);
+        CHECK(v[i]()->x == i);
```
AliceMazzoleni99, line 145: why CHECK twice

tests/scripts/test_free.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Print out the exception for confirmation
+    Log(LogLevel.INFOGREEN, f"✅ Test passed, exception was: {exc_info.value}")
+
+    assert str(exc_info.value) == "Shared memory is invalid or freed. Close resources before access."
+
+    cleanup(fp)
+
+
+
```
AliceMazzoleni99, line 93: Additional Test that should pass without throwing: 
`d = Detector()`
`d.hostname = "localhost"`
`d.hostname = "localhost"`
`d.exptime 1`

AliceMazzoleni99 (commented):
Looks good. Maybe add the additional test(should not thorw): 
`d = Detector()
d.hostname = "localhost"
d.hostname = "localhost"
d.exptime 1`

slsDetectorSoftware/src/Detector.cpp
thattil: It works. Done :+1:

slsDetectorSoftware/tests/test-SharedMemory.cpp
thattil: Maybe the view is updated? or I pushed in later? In any case, it does not anymore.

AliceMazzoleni99 (approved):

PR #1255, opened 2025-08-04
Dev/fix tests

thattil:
Fixed:
- in python: validate port < 1024 should throw
- validate port range: does not validate properly
- ctb image size calcualtion, was not including readout mode
- added REQUIRE_NOTHROW in the main tests in master_attributes test to be able to follow where it crashes
- update image size in ctb whenever dbit reorder, offset or list is updated in generalData
- also update the number of analog byutes and transceiver bytes in GeneralData as they are used in dataProcessor
- added an option to not test general tests.. off by default


python/slsdet/utils.py
```diff
@@ -281,6 +281,6 @@ def hostname_list(args):
 
 
 def validate_port(value):
-    if value <= 0 or value > 65535:
+    if value < 1024 or value > 65535:
         raise ValueError("port must be in range 1 - 65535")
```
AliceMazzoleni99, line 286: update error message - why cant we have port numbers < 1024?

tests/scripts/utils_for_test.py
```diff
@@ -252,7 +252,7 @@ def loadBasicSettings(name, d, fp):
     except Exception as e:
         raise RuntimeException(f'Could not load config for {name}. Error: {str(e)}') from e
     
-def ParseArguments(description, default_num_mods=1, markers=0):
+def ParseArguments(description, default_num_mods=1, markers=0, general_tests_option=False):
     parser = argparse.ArgumentParser(description)
```
AliceMazzoleni99, line 259: should markers be a boolean as well?

python/slsdet/utils.py
thattil: they are reserved. Updated the error message, thanks

AliceMazzoleni99 (approved):
Looks good. Added some minor remarks.

tests/scripts/utils_for_test.py
thattil: No, because you are giving a value to it.. for example, if you want to run only the [.cmdattr] for the master attributes test.. you would give 
-m "[.cmdattr]"

thattil: sorry, i misunderstood. The marker in the argument for ParseArgument should be boolean, I agree. Will fix it.

thattil: Done

PR #1256, opened 2025-08-05
dev/tests settingsdir relative path

thattil:
removed relative path compared to where executable run in test script for settingsdir


AliceMazzoleni99 (approved):
Perfect.

PR #1275, opened 2025-08-22
1000/verify shm

thattil:
- getUserDetails a free fucntion outside the Detector class to get user details directly from shared memory without creating a Detector object (detector constructor updates pid, user etc)
   - updated python bindings
   - updated command line
   - direct access to user command from cmdApp (does nto create Detector)
- removed verify, update arguments in creating shared memory (opening will always verify shared memory)
- refactored Detector::initsharedmemory( after opening detector shm, also open ctb if it exists)
- refactored Modle::initsharedmemory (split into create and open as here the function calls are obvious)
- if ctb shm exists when setting hostname, throw

#1276

### DIFF
--- a/slsDetectorSoftware/src/Detector.cpp
+++ b/slsDetectorSoftware/src/Detector.cpp
@@ -61,6 +61,54 @@ void freeSharedMemory(const int detectorIndex, const int moduleIndex) {
     }
 }
 
+std::string getUserDetails(const int detectorIndex) {
+
+    int numModules = 0;
+    defs::detectorType type = defs::GENERIC;
+    pid_t pid = -1;
+    std::string hostname = "Unknown";
+    std::string user = "Unknown";
+    std::string date = "Unknown";
+
+    SharedMemory<sharedDetector> detectorShm(detectorIndex, -1);
+    if (detectorShm.exists()) {
+        detectorShm.openSharedMemory(false);
+        if (detectorShm()->shmversion < DETECTOR_SHMAPIVERSION) {
+            detectorShm.unmapSharedMemory();
+            throw SharedMemoryError(
+                "Detector Shared memory version too old to get user details!");
+        }
+        numModules = detectorShm()->totalNumberOfModules;
+        type = detectorShm()->detType;
+        pid = detectorShm()->lastPID;
+        user = detectorShm()->lastUser;
+        date = detectorShm()->lastDate;
+
+        detectorShm.unmapSharedMemory();
+    }
+
+    for (int imod = 0; imod != numModules; ++imod) {
+        SharedMemory<sharedModule> moduleShm(detectorIndex, imod);
+        moduleShm.openSharedMemory(false);
+        if (moduleShm()->shmversion < MODULE_SHMAPIVERSION) {
+            LOG(logWARNING) << "Module Shared Memory too old to get hostname";
+        } else {
+            hostname = moduleShm()->hostname;
+        }
+        moduleShm.unmapSharedMemory();
+    }
+
+    std::ostringstream userDetails;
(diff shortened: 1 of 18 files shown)


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -61,6 +61,54 @@ void freeSharedMemory(const int detectorIndex, const int moduleIndex) {
...
+        pid = detectorShm()->lastPID;
+        user = detectorShm()->lastUser;
+        date = detectorShm()->lastDate;
+
+        detectorShm.unmapSharedMemory();
+    }
+
+    for (int imod = 0; imod != numModules; ++imod) {
+        SharedMemory<sharedModule> moduleShm(detectorIndex, imod);
+        moduleShm.openSharedMemory(false);
```
AliceMazzoleni99: Maybe check if they exist

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -61,6 +61,54 @@ void freeSharedMemory(const int detectorIndex, const int moduleIndex) {
...
+    }
+
+    std::ostringstream userDetails;
+    userDetails << "Detector Index: " << detectorIndex << "\n"
+                << "Number of Modules: " << numModules << "\n"
+                << "Type: " << type << "\n"
+                << "PID: " << pid << "\n"
+                << "User: " << user << "\n"
+                << "Date: " << date << "\n"
+                << "Hostname: " << hostname << "\n";
```
AliceMazzoleni99, line 116: Maybe easier for handling: Give an Info if no detector with detectorIndex exists in shared memory

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -210,10 +186,12 @@ void DetectorImpl::setHostname(const std::vector<std::string> &name) {
...
-        if (ctb_shm.exists())
-            ctb_shm.openSharedMemory(true);
-        else
-            ctb_shm.createSharedMemory();
+        if (ctb_shm.exists()) {
+            throw SharedMemoryError(
+                "This shared memory " + ctb_shm.getName() +
+                " should have been deleted before! Free it to continue.");
+        }
+        ctb_shm.createSharedMemory();
```
AliceMazzoleni99, line 194: I dont understand this part why cant it exist?

thattil: sethostname should only create the ctb shm (thats when it knows the type). Detector.cpp::sethostname already clears the shared memories. so should be clean as well.

slsDetectorSoftware/tests/Caller/test-Caller.cpp
```diff
@@ -3592,10 +3592,16 @@ TEST_CASE("frametime", "[.cmdcall]") {
...
+    std::ostringstream oss1, oss2, oss3;
+    caller.call("user", {}, -1, GET, oss1);
+    caller.call("user", {}, -1, GET, oss2);
+    caller.call("user", {}, -1, GET, oss3);
+    REQUIRE(oss1.str() == oss2.str());
+    REQUIRE(oss2.str() == oss3.str());
 
     // This is a get only command
-    REQUIRE_THROWS(caller.call("user", {}, -1, PUT));
+    // REQUIRE_THROWS(caller.call("user", {}, -1, PUT)); exit with failure
```
AliceMazzoleni99: should it be commented

thattil: removed all the caller tests.. just for help

AliceMazzoleni99 (approved):
Looks good.

PR #1278, opened 2025-08-25
1000/shm_fix_remove

thattil:
one doesnt need to open shared memory to call removesharedmemory, and calling hasMemoryvalid without opening will cause segfault (not used now, but could in the future)

### DIFF
--- a/slsDetectorSoftware/src/Detector.cpp
+++ b/slsDetectorSoftware/src/Detector.cpp
@@ -29,7 +29,6 @@ void freeSharedMemory(const int detectorIndex, const int moduleIndex) {
     if (moduleIndex >= 0) {
         SharedMemory<sharedModule> moduleShm(detectorIndex, moduleIndex);
         if (moduleShm.exists()) {
-            moduleShm.openSharedMemory(false);
             moduleShm.removeSharedMemory();
         }
         return;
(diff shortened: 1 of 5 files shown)


slsDetectorSoftware/src/Detector.cpp
```diff
@@ -48,15 +47,13 @@ void freeSharedMemory(const int detectorIndex, const int moduleIndex) {
     for (int i = 0; i < numDetectors; ++i) {
```
AliceMazzoleni99, line 47: Ive noticed There is one more open in line 44

AliceMazzoleni99: *41

AliceMazzoleni99 (changes requested):
Looks fine.

slsDetectorSoftware/src/Detector.cpp
thattil: but there, we want to open it to read the number of modules in shared memory (to delete the module shared memories)

AliceMazzoleni99: Oh true my fault.

AliceMazzoleni99 (approved):

PR #1282, opened 2025-08-25
1000/fix_actual_tests

thattil:
- fix acquire fail in tests (adcreg test)
- roi tests fail after overlapping invalid test and acquire after
- print udp dest mac in server properly
- fixed udp dst list get (server was not sending entry proper size to match proper struct size in client)
- updated server binaries and updated hard links in serverBin
- added documentation regarding gui:  zmqport and zmqip in terms of gui, rx_zmqstream


docs/src/configcommands.rst
```diff
@@ -201,6 +201,15 @@ Enabling the GUI automatically streams images from the receiver via ZMQ sockets.
 * Default: Receiver’s hostname (rx_hostname)
 
 
+.. note ::   
+
+    ``zmqport`` and ``zmqip`` need to be set up in shared memory before starting up the Gui. If the Gui is already running, change it in the Advanced Tab directly in the Gui.
+    
+    ``rx_zmqstream`` is set to 1 when the GUI is started, but has to be manually set back to 0 to disable zmq streaming in the receiver for better performance when not using the Gui.
+    
+    The GUI hwm values are set to a low number (from default library default of possibly 1000 to 2) to reduce the zmq buffer to display only some of the images. Resetting it back to -1 or 1000 should only matter if one plans to not use the Gui and still zmq stream without wanting to drop zmq images (eg. for external processing chain).
```
AliceMazzoleni99: type: default twice. 
I am not sure what hwm means?

slsDetectorSoftware/tests/Caller/test-Caller-rx.cpp
```diff
@@ -554,6 +556,12 @@ TEST_CASE("rx_roi", "[.cmdcall]") {
                                               "]]\n");
                 }
             }
+            // valid roi before acquiring
+            else {
+                std::ostringstream oss;
+                caller.call("rx_roi", {"10", "15"}, -1, PUT, oss);
+                REQUIRE(oss.str() == "rx_roi [[10, 15]]\n");
+            }
```
AliceMazzoleni99, line 564: Is this neccessary for the remainder of the test? E.g. create files for acquire?

slsDetectorSoftware/tests/Caller/test-Caller.cpp
```diff
@@ -2580,6 +2586,10 @@ TEST_CASE("numinterfaces", "[.cmdcall]") {
             caller.call("numinterfaces", {}, -1, GET, oss);
             REQUIRE(oss.str() == "numinterfaces 1\n");
         }
+        if (prev_val == 2 && det_type != defs::EIGER) {
+            det.setDestinationUDPList({prev_udp_dest}, 0);
+            det.setSourceUDPIP2({prev_src_ip2}, {0});
+        }
```
AliceMazzoleni99, line 2592: why is this relevant? MInor: Check for Eiger is unnecessary

AliceMazzoleni99 (commented):
Reviewed it. Can it be that the ctbDetectorServer and jungfrauDetectorServer binaries are missing?

slsDetectorSoftware/tests/Caller/test-Caller-rx.cpp
thattil: Yes. because it could fail while trying to write the master files (writing each master attribute)

slsDetectorSoftware/tests/Caller/test-Caller.cpp
thattil: It is needed because Eiger does not have udp_destip2, but has udp_dstport2

thattil: And it is relevant. Setting number of interfaces to 1 wipes out the udp details of the second interface on the server.  So, one has to set it up again since we are just testing here and we mgiht still need it for other tests further down the line.

PR #1303, opened 2025-09-12
Pattern unification & Matterhorn Changes

### LINKED ISSUE #1144: ctb: check half empty instead of not empty flag before reading n samples (a frame)

mmarti04:


mmarti04:
@thattil i tested your rewrite of Xilinx_pll.c again and it works now. 
From my side this is ready to merge.

PR #1306, opened 2025-09-17
Dev/include file

### LINKED ISSUE #1304: ctb: define
This is to handle all the hardcoded addresses and bits  in reg command for ctb.
 
1. define:We will specify 'addr' and 'bit' for the define command and add a deflinelist command to get the list.
    sls_detector_put   define   addr   MATTERHORN_CTRL    0x202
    sls_detector_put   define   bit    run_busy   31

2. Using define:
These defines can then be used with reg, setbit, getbit and clearbit.
    sls_detector_put    reg     MATTERHORN_CTRL      (run_busy | start_exp)
    sls_detecot_put   clearbit  MATTERHIRN_CTRL   start_exp

3. definelist:
One can get the list of defines using 'definelist'
    sls_detector_get   definelist    addr

4. include:
We already have 'parameters' that takes a file and executes a 'put' for every line. (Difference between 'parameters' and 'config' is that 'config' frees shared memory beforehand. We will create an alias 'include' for 'parameters', which seems to be apt. 

Suggestion is that both 1. and 2. be  separated into 2 files and then "included" in the config file for better encapsulation and reusability. This of course will not be enforced, but is just suggested as a good practice.

5. Ctb Shared memory:
This info will be put in the ctb shared memory. And since only the ctb shared memory changes, it doesnt become a major release. 🙂
 @mmarti04

thattil:
same as parameters command in command line and python


erikfrojdh:
@thattil Getting errors after adding colorama. Will try to debug but maybe we have to take out the integration tests.

erikfrojdh (approved):
LGTM

python/slsdet/detector.py
```diff
@@ -147,6 +147,33 @@ def parameters(self, value):
             value = ut.make_string_path(value)
         self.loadParameters(value)
 
+
+    @property
+    def include(self):
```
erikfrojdh, line 154: I'm having second thoughts about the include in python. Not sure this looks intuitive... loadParameters is maybe better on a detector object, but we can discuss. On the other hand it's consistent with `d.parameters`

thattil: We are merging so the others can use include in the config file already. We can make a separate Pr to remove this include in python.

PR #1307, opened 2025-09-18
ctb: add patternstart command, xilinx: fix frequency

### LINKED ISSUE #1305: xilinx and ctb: patternstart command, xilinx: fix frequency
on both ctb versions each pattern sends a packet right now. We should have a command which runs the pattern without sending packets. (As it is done in mythen)

- also fix frequency in xilinx ctb server ( from PR #1303)

mmarti04:
added patternstart command for CTB. 
It will block end of pattern execution udp packets which are usually send by p start.
This introduces an easy way to avoid nonsense data packets in the output file.
Right now the acquired data for Matterhorn has to be filtered afterwards into udp packets from readout patterns and nonsense udp packets from other patterns. 

frixes for PR (#1303)
- removed tick clk => syncclk
- using clkFrequency and setting to defaults in setup_detctor
- setting thsi variable in getFrequcny to get exact value
- xilinx: set frequency returns ok or fail
- xilinx frequency in kHz defaults
#1309


slsDetectorServers/slsDetectorServer/src/loadPattern.c
```diff
@@ -634,6 +634,31 @@ void startPattern() {
     LOG(logINFOBLUE, ("Pattern done\n"));
 }
 #endif
+#ifdef CHIPTESTBOARDD
```
thattil: use #elif with function definiton and the print for start pattern and done outside.

slsDetectorServers/ctbDetectorServer/RegisterDefs.h
```diff
@@ -384,6 +384,13 @@
...
 
+/* Streaming Control RW regiser */
+#define STREAMING_CTRL_REG         (0x3D << MEM_MAP_SHIFT)
+#define STREAMING_CTRL_ENA_OFST    (15)
+#define STREAMING_CTRL_ENA_MSK     (0x00000001 << STREAMING_CTRL_ENA_OFST)
+#define STREAMING_CTRL_SELECT_OFST (0)
+#define STREAMING_CTRL_SELECT_MSK  (0x0000003F << STREAMING_CTRL_SELECT_OFST)
+
 /* Pattern Control RW register */
 #define PATTERN_CNTRL_REG (0x88 << MEM_MAP_SHIFT)
```
thattil, line 396: This server has been compiled without updating the versions file

docs/src/pattern.rst
```diff
@@ -119,4 +119,12 @@ DIO: Driving the 32 FPGA pins corresponding to the lowest 32 bits of the patioct
...
 
-For Mythen3 the pattern word only connects to output pins of the FPGA when the pattern is running. Afterwards the signals will switch back to other logic in the FPGA. Both CTB's hold the last executed pattern word until a new pattern is started.
\ No newline at end of file
+For Mythen3 the pattern word only connects to output pins of the FPGA when the pattern is running. Afterwards the signals will switch back to other logic in the FPGA. Both CTB's hold the last executed pattern word until a new pattern is started.
+
+**Relation of received data to pattern execution**
+
+In the default configuration the Ctb will send out udp packets to the sls_receiver for every end of a pattern execution. This behavior can be changed using STREAMING_CTRL_REG, where one can configure a bit position in the 64-bit pattern word to trigger udp packets. This allows to send more than one packet per pattern or also no packets at all.
+
+The "patternstart" command on the Ctb exists because people do not read this documentation and do not expect every pattern to send packets. It automatically suppresses the packet at the end of a pattern. Additional packets via STREAMING_CTRL_REG are still possible.
```
thattil: I would remove this line about what users expect and jsut say what it does. "command on the ctb exectues the pattern and automatically suppresses the udp packet generation at the end of the pattern.".. Additional packets via .. does this also work with patternstart? IF so, I would change to "command on the ctb exectues the pattern. As long as streaming_ctrl_reg is disabeld (or something similar), every pattern execution using this command will not generate/send UDP packets"

slsDetectorServers/slsDetectorServer/src/loadPattern.c
```diff
@@ -634,6 +634,31 @@ void startPattern() {
...
+void startPattern() {
+    // we only want to run the pattern here. No acquisition, no UDP packets
+    
+    // disable 10G UDP temporarily
+    // except if the pattern explicitly contains udp trigger points
+    uint32_t conf_reg_tmp = bus_r(CONFIG_REG);
+    if((bus_r(STREAMING_CTRL_REG) & STREAMING_CTRL_ENA_MSK) == 0){
```
thattil: and if its 10g enable?

thattil (changes requested):

slsDetectorServers/slsDetectorServer/src/loadPattern.c
mmarti04: then it is disabled here and enabled again in line 658. That's how the packed is blocked in the firmware.

thattil:
Ive changed the way run_clk is used. I see that clkFrequency was defined but not used anywhere. It seems to resolve the test failure on period.  But what is happening with tick_clk?

PR #1312, opened 2025-09-29
Dev/define cmd

### LINKED ISSUE #1304: ctb: define

thattil:
New:
- commands: define_reg, define_bit, definelist_reg, definelist_bit

Definitions can only be 32 characters long and is stored in shared memory only for ctb and xilinx ctb.

Define_reg: Defines a register address
   - examples:
              sls_detector_put define_reg test_reg 0x200
              sls_detector_get define_reg test_reg # => 0x200
              sls_detector_get define_reg 0x200 # => test_reg

              sls_detector_put reg test_reg 0x1
              sls_detector_put setbit test_reg 2
              sls_detector_put clearbit test_reg 2
              sls_detector_get getbit test_reg 2

Define_bit: Defines a bit position and register address together
- examples:
              sls_detector_put define_bit test_bit test_reg 2
              sls_detector_put define_bit test_bit 0x200 2

              sls_detector_get define_bit test_bit # => test_reg 2
              sls_detector_get define_bit test_reg 2 # =>test_bit
              sls_detector_get define_bit 0x200 2 # => test_bit

              sls_detector_put setbit test_bit
              sls_detector_put clearbit test_bit
              sls_detector_get getbit test_bit

All of them also support values as before. 

- Had to move these from commands.yaml to cmdspecial to be able to allow parsing by string (defined names) or integer (original method).

address and bit names are only for ctb and xilinx=> written to CtbConfig shared memory. Change shm date.
API added to set them and get them, and get the list.


slsDetectorSoftware/include/sls/Detector.h
```diff
@@ -1837,6 +1837,53 @@ class Detector {
...
+
+    /** [CTB][Xilinx CTB] */
+    int getBitDefinitionByName(const std::string &name) const;
+
+    /** [CTB][Xilinx CTB] */
+    void clearBitDefinitions();
+
+    /** [CTB][Xilinx CTB] */
+    void
+    setBitDefinitions(const std::vector<std::pair<std::string, int>> &list);
```
erikfrojdh: same as the above a std::map<string, int>?

slsDetectorSoftware/include/sls/Detector.h
```diff
@@ -1837,6 +1837,53 @@ class Detector {
...
+
+    /** [CTB][Xilinx CTB] */
+    void clearRegisterDefinitions();
+
+    /** [CTB][Xilinx CTB] */
+    void setRegisterDefinitions(
+        const std::vector<std::pair<std::string, int>> &list);
+
+    /** [CTB][Xilinx CTB] */
+    std::vector<std::pair<std::string, int>> getRegisterDefinitions() const;
```
erikfrojdh: Vector or std::map<string, int>

slsDetectorSoftware/src/CtbConfig.cpp
```diff
@@ -238,4 +238,176 @@ std::vector<std::string> CtbConfig::getSlowADCNames() const {
 
 const char *CtbConfig::shm_tag() { return shm_tag_; }
 
+std::optional<Entry *> CtbConfig::findEntryByName(const std::string &name,
```
erikfrojdh: My gut feeling is that the lookup would be much easier with map/unordered map

erikfrojdh (commented):
Think about the implementation.

slsDetectorSoftware/include/sls/Detector.h
thattil: Will do

slsDetectorSoftware/include/sls/Detector.h
thattil: i guess you mean map? cant do vector<string, int>. Ill implement that anyway.

python/slsdet/utils.py
```diff
@@ -141,21 +141,38 @@ def make_ip(arg):
...
-        return {key: transform(value) for key, value in arg.items()}
+        return {key: _make(value, transform) for key, value in arg.items()}
     elif isinstance(arg, list):
-        return [transform(a) for a in arg]
+        return [_make(a, transform) for a in arg]
     elif isinstance(arg, tuple):
-        return tuple(transform(a) for a in arg)
+        # special case for BitPosition
+        if transform is _slsdet.BitPosition:
+            addr, bit = arg
```
AliceMazzoleni99: why only for transform _slsdet.BitPosition?

python/src/detector.cpp
```diff
@@ -1969,29 +2045,55 @@ void init_det(py::module &m) {
...
         "writeRegister",
         (void (Detector::*)(uint32_t, uint32_t, bool, sls::Positions)) &
             Detector::writeRegister,
         py::arg(), py::arg(), py::arg() = false, py::arg() = Positions{});
+    CppDetectorApi.def(
+        "setBit",
+        (void (Detector::*)(sls::BitPosition, bool, sls::Positions)) &
+            Detector::setBit,
+        py::arg(), py::arg() = false, py::arg() = Positions{});
     CppDetectorApi.def(
```
AliceMazzoleni99, line 2113: Do we want to enable both strong types and the non strong types?

python/src/detector_in.cpp
```diff
@@ -5,8 +5,10 @@
 #include "sls/Detector.h"
 #include "sls/ToString.h"
 #include "sls/network_utils.h"
+#include "sls/bit_utils.h"
 #include "sls/sls_detector_defs.h"
 #include "sls/TimeHelper.h"
+#include "sls/bit_utils.h"
```
AliceMazzoleni99, line 11: only keep once

python/slsdet/utils.py
thattil: Because Bit Position has 2 values unlike the others..

python/src/detector.cpp
thattil: In Detector.h, we are keeping both to make it backward compatible. Did we decide this in a prior meeting? If not, maybe we should discuss this. Python bindings if it should stay the same as Detector class. We should also discuss if it hasnt been.

python/src/detector_in.cpp
thattil: Done!

slsDetectorSoftware/generator/commands.yaml
```diff
@@ -2761,6 +2761,66 @@ sleep:
             - argc: 2
               arg_types: [ int, special::time_unit ]
```
AliceMazzoleni99, line 2763: what is this used for?

AliceMazzoleni99: is there no help message?

slsDetectorSoftware/generator/extended_commands.yaml
AliceMazzoleni99, line 1: I didnt look over this part

slsDetectorSoftware/generator/commands.yaml
thattil: I think this is used for the infer script to infer if the command is for sls_detector_put or sls_detector_get. This command (and all others with "is_description: true") is implemented in CallerSpecial.cpp. And writing it here only helps to infer the put or get. This is what I understand.

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+        throw RuntimeError("Cannot use define at module level. Use the default "
+                           "multi-module level");
+    }
+
+    // mode: addr or bit
+    if (args.empty()) {
+        WrongNumberOfParameters(2);
+    }
+    std::string mode = args[0];
+    if (mode == "addr") {
```
AliceMazzoleni99: is it addr or define addr not clear from helper string

thattil: 'define' is the command.. 'addr' or 'bit' is the mode.. There are examples in the helper string.

\n\tsls_detector_put define addr test_reg 0x200"
              "\n\tsls_detector_put define bit test_bit test_reg 2"
              "\n\tsls_detector_put define bit test_bit 0x200 2"

And many more. Maybe we can discuss tomorrow how I can improve this to be more readable. 

I also see that I didnt update the PR description to reflect the change to "addr" naming. Ill fix this now.

slsDetectorSoftware/generator/extended_commands.yaml
thattil: The extended commands.yaml you dont really need to, as it is auto generated. Upto you.

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+        argsSize -= 1; // last arg is --validate
+    }
+    std::string addr_or_bitname = args[0];
+
+    // bit name
+    if (argsSize == 1) {
+        auto det_type = det->getDetectorType().squash();
+        if (det_type != defs::XILINX_CHIPTESTBOARD &&
+            det_type != defs::CHIPTESTBOARD) {
+            throw RuntimeError("Could not parse bit name " + addr_or_bitname +
```
AliceMazzoleni99: why is this error message only here?

python/slsdet/utils.py
```diff
@@ -141,21 +141,38 @@ def make_ip(arg):
...
     return _make(arg, _slsdet.MacAddr)
 
+def make_register_address(arg):
+    return _make(arg, _slsdet.RegisterAddress)
+
+def make_bit_position(arg):
+    return _make(arg, _slsdet.BitPosition)
+
+def make_register_value(arg):
+    return _make(arg, _slsdet.RegisterValue)
```
AliceMazzoleni99: use it in detector.py

python/slsdet/utils.py
AliceMazzoleni99: ah got it.

python/slsdet/utils.py
AliceMazzoleni99: Or actually its never used right now I dont think we use it anyway.

slsDetectorSoftware/src/CallerSpecial.cpp
AliceMazzoleni99: Maybe cretae two functions define_reg and define_bit

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+    if (action == defs::HELP_ACTION) {
+        os << "[Ctb][Xilinx Ctb]"
+              "\n\t[addr] [addr name] [address]"
+              "\n\t[bit] [bit name] [bit position] [addr name/ address]"
+              "\n\n\tSets a user defined register or bit name in shared "
+              "memory. The name can be upto 32 characters long."
+              "\n\teg."
+              "\n\tsls_detector_put define addr test_reg 0x200"
+              "\n\tsls_detector_put define bit test_bit test_reg 2"
+              "\n\tsls_detector_put define bit test_bit 0x200 2"
```
AliceMazzoleni99: as this can still be mixed up add an info Defining test_bit as bit 2 of register with address 0x200

AliceMazzoleni99: Maybe put it in Detector::setBitPosition then the info is also shown when called from python

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+            // get name from address
+            if (is_hex_or_dec_uint(args[1])) {
+                auto addr = RegisterAddress(args[1]);
+                auto t = det->getRegisterDefinition(addr);
+                os << t << '\n';
+            }
+            // get address from name
+            else {
+                auto t = det->getRegisterDefinition(args[1]);
+                os << t.str() << '\n';
```
AliceMazzoleni99: mmh i think one should return the same thing twice. e.g. Detector::getRegisterDefinition always returns a std::map<std::string, RegisterAdress>

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+              "validation when writing to it.\n\t[Eiger] +0x100 for only left, "
+              "+0x200 for only right.\n\t\t[Ctb][Xilinx_Ctb] Address can also "
+              "be a user-defined name that was set previously using the define "
+              "command. Value can be a user-defined bit name as well or "
+              "combined with '|' without spaces eg. 'test_bit|run_bit' .When "
+              "using bit names, avoid register name/ address as the bit name "
+              "is tied to a specific register already.\n"
+              "\n\n\teg."
+              "\n\tsls_detector_get reg test_reg"
+              "\n\tsls_detector_put reg test_bit"
```
AliceMazzoleni99: should this be sls_detector_get

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+    }
+
+    // address and value
+    if (argsSize == 2) {
+        std::string saddr = args[0];
+        std::string sval = args[1];
+        if (!is_hex_or_dec_uint(sval)) {
+            throw RuntimeError(
+                "If " + sval +
+                " is a bit name, use only one argument to specify it.");
```
AliceMazzoleni99: why bit name? Dont get the error message

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+        }
+        return std::make_pair(parseAddress(saddr), RegisterValue(sval));
+    }
+    // parse bit name (or concatenation of them)
+    else if (argsSize == 1) {
+        std::string sbitNames = args[0];
```
AliceMazzoleni99: is it the register or bit?

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+    return os.str();
+}
+
+std::string Caller::reg(int action) {
+    std::ostringstream os;
+    if (action == defs::HELP_ACTION) {
+        os << "[address] [32 bit "
+              "value][(optional)--validate]\n\t[Mythen3][Gotthard2] "
+              "Reads/writes to a 32 bit register in hex. Advanced "
+              "Function!\n\tGoes to stop server. Hence, can be called while "
```
AliceMazzoleni99: MMh maybe I would completely seperate bit names from this command and only use it in bitoperations

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+        if (det_type != defs::XILINX_CHIPTESTBOARD &&
+            det_type != defs::CHIPTESTBOARD) {
+            throw RuntimeError("Could not parse reg value " + sbitNames +
+                               ". User defined bit definitions only supported "
+                               "for ctb and xilinx_ctb. Use an actual hard "
+                               "coded value for this detector.");
+        }
+
+        // if args0 does not have '|', then get bit definition directly
+        if (sbitNames.find('|') == std::string::npos) {
```
AliceMazzoleni99: Here again I would seperate bitnames and implement those things in parseBitposition

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1436,5 +1436,446 @@ std::string Caller::sleep(int action) {
...
+              "\n\tsls_detector_put setbit test_bit"
+              "\n\tsls_detector_put clearbit test_bit";
+        os << '\n';
+    } else {
+        if (action != defs::GET_ACTION && action != defs::PUT_ACTION) {
+            throw RuntimeError("Unknown action");
+        }
+
+        auto validate = parseValidate();
+        auto bitPosition = parseBitPosition(validate);
```
AliceMazzoleni99: again add an INFO if no register definition is used that you set the value at register {adress} to {value}

python/src/detector.cpp
```diff
@@ -1796,6 +1800,78 @@ void init_det(py::module &m) {
...
+    CppDetectorApi.def(
+        "getBitDefinition",
+        (sls::BitPosition(Detector::*)(const std::string &) const) &
+            Detector::getBitDefinition,
+        py::arg());
+    CppDetectorApi.def("getBitDefinition",
+                       (std::string(Detector::*)(sls::BitPosition) const) &
+                           Detector::getBitDefinition,
+                       py::arg());
+    CppDetectorApi.def("clearBitDefinitions",
```
AliceMazzoleni99, line 1868: All the bit operations cant be called as a bit definition name from python. Maybe overload these in detector.py and pass a string e.g. bit name and also declare a dictionary with bit name and value. getting an item then calls getBitDefinition() and then clearBit or setBit depending on the set value.

slsDetectorSoftware/src/CtbConfig.h
```diff
@@ -1,10 +1,61 @@
...
+
 namespace sls {
 
-#define CTB_SHMAPIVERSION 0x250820
-#define CTB_SHMVERSION    0x250820
+#define CTB_SHMAPIVERSION 0x251031
+#define CTB_SHMVERSION    0x251031
+
+#define CTB_NAME_LENGTH 32
+
```
AliceMazzoleni99: maybe a std::map<std::string, RegisterAdress> as registers member would have been easier

AliceMazzoleni99: Probably just dont see why it doesnt work

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -2407,37 +2402,24 @@ defs::dacIndex Detector::getDacIndex(const std::string &name) const {
 }
 
 void Detector::setDacName(const defs::dacIndex i, const std::string &name) {
-    auto dettype = getDetectorType().squash();
-    if (dettype != defs::CHIPTESTBOARD && dettype != defs::XILINX_CHIPTESTBOARD)
-        throw RuntimeError("Named dacs only for CTB");
     pimpl->setCtbDacName(i, name);
```
AliceMazzoleni99, line 2405: it works for all detectors?

AliceMazzoleni99: mmh maybe stay consistent and only test for isChipTestBoard on DetectorImplementation side. Can be another PR

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -2566,19 +2502,85 @@ defs::dacIndex Detector::getSlowADCIndex(const std::string &name) const {
...
+}
+
+bool Detector::hasRegisterDefinition(RegisterAddress addr) const {
+    return pimpl->hasRegisterDefinition(addr);
+}
+
+RegisterAddress Detector::getRegisterDefinition(const std::string &name) const {
+    return pimpl->getRegisterDefinition(name);
+}
+
```
AliceMazzoleni99, line 2532: maybe these should return a RegisterDefinition or a std::pair<std::string, RegisterAddress> if we use std::unordered_map

slsDetectorSoftware/src/DetectorImpl.cpp
```diff
@@ -1935,86 +1935,240 @@ void DetectorImpl::setBadChannels(const std::vector<int> list, Positions pos) {
...
+        throw RuntimeError("Bit Definitions only for CTB");
+    return ctb_shm()->hasBitPosition(bitPos);
+}
+
+BitPosition DetectorImpl::getBitDefinition(const std::string &name) const {
+    if (!isChipTestBoard())
+        throw RuntimeError("Bit Definitions only for CTB");
+    return ctb_shm()->getBitPosition(name);
+}
+
```
AliceMazzoleni99, line 2153: same here maybe return BitDefinition for both

slsDetectorSoftware/src/inferAction.cpp
```diff
@@ -788,6 +788,34 @@ int InferAction::defaultpattern() {
     }
 }
 
+int InferAction::define() {
+
+    if (args.size() == 2) {
```
AliceMazzoleni99, line 801: cant there be more than 2 arguments e.g. define bit register_adress bit_position

python/src/detector.cpp
thattil: We keep deprecated functions (not strong typed arguments) in Detector cpp class until a major version where we would break backwards compatibility. About the warning in python bindings because of it, we still need to decide how to handle it.

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: Why should it return a map when there is only a single value required here... I believe if one needs to just get a single registeraddress from a name, then it needn't be more complex and just return the string. Or have I misunderstood you?

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: but this could also discourage hard coded values into a register. Bit names concatenation would give a clearer idea of what is being set in this register.

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: No, setting just the bit will set only that bit in that register.

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: Will add at command line level.

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: I didnt understand this question. This error message is if you try to use bit name or address name definitions as they are only for ctb and xilinx ctb. They are in other places as well.

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: If you give 2 arguments.. 0x200 test_bit... test_bit is already connected to a register.. so you need to give only test_bit..

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: They are just bit names concatenated. You dont need a register address as all bit names are connected to a register address. I jsut ensure all these addresses are the same, set the bits and set that value to the register.

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: parse bit position is only parsing the combination of address name/value + bit position or a bit name.

This concatenation of bit names is only an alternative for writing a value to a register. It is not to parse what the bit position is.

slsDetectorSoftware/src/CtbConfig.h
thattil: I cant use that in shared memory because it allocates dynamically.

slsDetectorSoftware/src/Detector.cpp
thattil: It does. I just moved it to DetectorImpl to hide the implementation details as it anyway calls DetectorImpl. Also because I have more of these checks for the define in DetectorImpl.

slsDetectorSoftware/src/Detector.cpp
thattil: I dont see why it has to be more complicated. It only requires a string  as return type.

slsDetectorSoftware/src/DetectorImpl.cpp
thattil: I am not sure I agree. BitPosition is simpler. Besides Bit definition is only defined internally for the shared memory structure.

slsDetectorSoftware/src/inferAction.cpp
thattil: Yes there can be. However, if there are only 2 arguments, it can only be a get.

slsDetectorSoftware/src/inferAction.cpp
```diff
@@ -788,6 +788,34 @@ int InferAction::defaultpattern() {
...
+        return slsDetectorDefs::GET_ACTION;
+    }
+
+    if (args.size() == 3) {
+        return slsDetectorDefs::PUT_ACTION;
```
thattil, line 806: But this is incorrect. sls_detector_get define bit 0x200 20. This can be a get. I have to fix this.

slsDetectorSoftware/src/CallerSpecial.cpp
AliceMazzoleni99: But why is this in parseRegisterAdress and not parseBitPoistion? Maybe I misunderstand what parseRgeisterAdress should do. Or when you call reg["bit_name"] does it give you the value at the bit poistion? or do you have to use getbit?

slsDetectorSoftware/src/CtbConfig.h
AliceMazzoleni99: Ah oke got it.

python/src/detector.cpp
thattil: You can using d.setBit(d.getBitDefinition('start_trigger_bit')). However, tomorrow afternoon if we decide to implement "define" property in python, this can be implemented directly as well.

python/slsdet/utils.py
thattil: I think we discussed this yesterday. That when a user uses it, they should use it to convert it to RegisterValue to use the non deprecated functions.  Right now, both are possible.

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: It is similar but the name is not "parseRegisterAddress". This function only passes a register address. 

However, the method we are discussing is 'parseRegisterAddressAndValue'. It is called when you "write" a value to register. This can currently be either a 
- combination of an address (value or name) and a hardcoded value, or
- concatenation of bit names (here you do not require an address as bit names are already connected to an address)

slsDetectorSoftware/src/CallerSpecial.cpp
AliceMazzoleni99: Oke got it. Sorry misunderstood.

slsDetectorSoftware/src/Detector.cpp
AliceMazzoleni99: I see. For me it is just more explicit if one exposes getRegisterAdress or getRegisterName to Detector and only use those. I found it at first a bit confusing that the funtion returns different things for different input values. But its only a minor suggestion.

thattil: That makes sense actually. The issue is with BitPosition. getBitPosition().bitPosition(). Lets discuss this tomorrow and see what is more intuitive.

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+
+    constexpr bool operator==(const RegisterAddress &other) const {
+        return (addr_ == other.addr_);
+    }
+    constexpr bool operator!=(const RegisterAddress &other) const {
+        return (addr_ != other.addr_);
+    }
+} __attribute__((packed));
+
+class BitPosition {
```
erikfrojdh: Are we happy with the BitPosition name? To me it's not intuitive that it takes an address and position. Maybe BitDefinition/BitAddress or something similar?

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -3,6 +3,7 @@
 #pragma once
 
 #include <bitset>
+#include <stdint.h>
```
erikfrojdh: Please use `#include <cstdint>` which is the C++ version of the header

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+    std::string str() const;
+    operator uint32_t() const noexcept { return addr_; }
+
+    constexpr bool operator==(const RegisterAddress &other) const {
+        return (addr_ == other.addr_);
+    }
+    constexpr bool operator!=(const RegisterAddress &other) const {
+        return (addr_ != other.addr_);
+    }
+} __attribute__((packed));
```
erikfrojdh: Do you have a specific reason to use attribute packed? With a single member it doesn't make a different except that it changes the alignment from 4 to 1 bytes.

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+    void setAddress(RegisterAddress address) noexcept { addr_ = address; }
+    void setBitPosition(int bitPosition) noexcept { bitPos_ = bitPosition; }
+
+    constexpr bool operator==(const BitPosition &other) const {
+        return (addr_ == other.addr_ && bitPos_ == other.bitPos_);
+    }
+    constexpr bool operator!=(const BitPosition &other) const {
+        return !(*this == other && bitPos_ == other.bitPos_);
+    }
+} __attribute__((packed));
```
erikfrojdh: Same thing here. With two 4 byte members there should be no gaps and I don't think attribute packed is needed.

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+    }
+    constexpr bool operator!=(const RegisterAddress &other) const {
+        return (addr_ != other.addr_);
+    }
+} __attribute__((packed));
+
+class BitPosition {
+  private:
+    RegisterAddress addr_{0};
+    int bitPos_{0};
```
erikfrojdh: Since we use uint32_t in RegisterAddress I would use int32_t for consistency.

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+    RegisterAddress address() const noexcept { return addr_; }
+    int bitPosition() const noexcept { return bitPos_; }
+    void setAddress(RegisterAddress address) noexcept { addr_ = address; }
+    void setBitPosition(int bitPosition) noexcept { bitPos_ = bitPosition; }
+
+    constexpr bool operator==(const BitPosition &other) const {
+        return (addr_ == other.addr_ && bitPos_ == other.bitPos_);
+    }
+    constexpr bool operator!=(const BitPosition &other) const {
+        return !(*this == other && bitPos_ == other.bitPos_);
```
erikfrojdh: You can simplify it to: `return !(*this == other);`

slsSupportLib/tests/test-bit_utils.cpp
```diff
@@ -42,4 +43,141 @@ TEST_CASE("Get set bits from 523") {
...
+    // ensure correct exception message
+    try {
+        BitPosition(RegisterAddress(0x305), 32);
+    } catch (const std::exception &e) {
+        REQUIRE(std::string(e.what()) == "Bit position must be between 0 and 31.");
+    }
+
+}
+
+TEST_CASE("Implicitly convert to uint for sending over network", "[support][bit_utils]") {
```
erikfrojdh: I'm a bit lost. Why would we want to convert it to an uint64 and then send it instead of sending the address as a 32 bit value directly?

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+  private:
+    uint32_t value_{0};
+
+  public:
+    constexpr RegisterValue() noexcept = default;
+    explicit constexpr RegisterValue(uint32_t value) noexcept : value_(value) {}
+    explicit RegisterValue(const std::string &value);
+
+    std::string str() const;
+    operator uint32_t() const noexcept { return value_; }
```
erikfrojdh: I'm not sure about the conversion operator... I would prefer a `uint32_t RegisterValue::value()` instead. To easy conversion defeats the purpose of having a strong type.

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
     return set_bits;
 }
+
+class RegisterAddress {
+  private:
+    uint32_t addr_{0};
+
+  public:
+    constexpr RegisterAddress() noexcept = default;
+    explicit RegisterAddress(uint32_t address) : addr_(address) {}
```
erikfrojdh: Can be constexpr

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+
+  public:
+    constexpr RegisterValue() noexcept = default;
+    explicit constexpr RegisterValue(uint32_t value) noexcept : value_(value) {}
+    explicit RegisterValue(const std::string &value);
+
+    std::string str() const;
+    operator uint32_t() const noexcept { return value_; }
+
+    RegisterValue &operator|=(uint32_t rhs) noexcept {
```
erikfrojdh: Do we want to allow this with an uint32_t and not only with other RegisterValues?

slsSupportLib/include/sls/bit_utils.h
```diff
@@ -18,4 +19,76 @@ template <typename T> std::vector<int> getSetBits(T val) {
...
+    int bitPos_{0};
+
+  public:
+    constexpr BitPosition() noexcept = default;
+    BitPosition(RegisterAddress address, int bitPosition);
+    std::string str() const;
+
+    RegisterAddress address() const noexcept { return addr_; }
+    int bitPosition() const noexcept { return bitPos_; }
+    void setAddress(RegisterAddress address) noexcept { addr_ = address; }
```
erikfrojdh: Is there a need for the setters? Or we just do it on construction?

erikfrojdh (changes requested):
Added my comments. Mostly very quick fixes.  I have an idea how to make the python part a bit easier to use but will try tomorrow.

erikfrojdh:
This could be used in python:
```python
In [17]: regs = types.SimpleNamespace(**d.getRegisterDefinitions())

In [18]: d.reg[regs.something]
Out[18]: [0x0]

```

python/src/bit.cpp
```diff
@@ -0,0 +1,55 @@
...
+using sls::BitPosition;
+
+void init_bit(py::module &m) {
+
+    py::class_<RegisterAddress>(m, "RegisterAddress")
+        .def(py::init())
+        .def(py::init<const std::string &>())
+        .def(py::init<uint32_t>())
+        .def(py::init<const RegisterAddress &>())
+        .def("__repr__", &RegisterAddress::str)
```
erikfrojdh: For __repr__ do we want it to print RegisterAddress(0x45)? str should probably be only 0x45 for writing to file but the type might be useful when doing interactive stuff

slsDetectorSoftware/src/CtbConfig.h
```diff
@@ -1,10 +1,61 @@
...
+class RegisterDefinition {
+  private:
+    char name_[CTB_NAME_LENGTH]{};
```
erikfrojdh: Why use a macro and not constexpr static?

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: Makes sense. Its not intuitive. Will remove bit names or concatenation of bit names from the reg command

slsDetectorSoftware/src/CallerSpecial.cpp
thattil: Removing bit namees and concatenation in reg command to avoid misinterpretations. So this function will be removed.

slsSupportLib/include/sls/bit_utils.h
thattil: I didnt really get this? We use uint32_t for RegisterAddress and here we use RegisterAddress, so it should be the same right?

slsSupportLib/include/sls/bit_utils.h
thattil: To Modify it? I have only used it  once in a tests though.. we can remove it if its unnecessary..

slsSupportLib/include/sls/bit_utils.h
thattil: Ah Makes sense.

slsSupportLib/include/sls/bit_utils.h
thattil: Done. Allowed to concatenate with both RegisterValue and uint32_t

slsSupportLib/include/sls/bit_utils.h
AliceMazzoleni99: I think its because the bitPos can be a value between 0-31 uint8_t makes more sense or why int32_t?

slsDetectorSoftware/src/CtbConfig.h
thattil: Using it for SHMVERSION. For CTB_NAME_LENGTH suing inling constexpr at namespace level as it is used in multiple classes.

slsSupportLib/include/sls/bit_utils.h
thattil: ah about bitPosition. Got it. Will change to uint32_t as well.

thattil: Done

slsSupportLib/include/sls/bit_utils.h
thattil: RegisterAddresss and RegisterValue: Removed the implicit conversions.
RegisterAddress: Changed member name from address_ to value_ and method as well to value(). 
RegisterValue: Also added | operator to be able to concatenate with uint32_t. Same in python bindings (but could not find the tests to modify).

python/slsdet/utils.py
```diff
@@ -141,21 +141,38 @@ def make_ip(arg):
 def make_mac(arg):
     return _make(arg, _slsdet.MacAddr)
 
+def make_register_address(arg):
+    return _make(arg, _slsdet.RegisterAddress)
```
AliceMazzoleni99: maybe add this in __init__ file e.g. from .utils import make_register_address otherwise people will for sure not find it and use the struct

slsSupportLib/tests/test-bit_utils.cpp
thattil: Implicit conversion removed

slsSupportLib/include/sls/bit_utils.h
thattil: Removed.

python/slsdet/registers.py
```diff
@@ -1,13 +1,28 @@
...
+from .utils import element
+
 class Register:
     def __init__(self, detector):
         self._detector = detector
 
+    @element
     def __getitem__(self, key):
+        if isinstance(key, str):
+            key = self._detector.getRegisterDefinition(key)
```
erikfrojdh: Needs to be updated with `getRegisterDefinitionAddress`

python/slsdet/registers.py
```diff
@@ -1,13 +1,28 @@
...
         return self._detector.readRegister(key)
 
     def __setitem__(self, key, value):
+        if isinstance(key, str):
+            key = self._detector.getRegisterDefinition(key)
```
erikfrojdh: Here as well: `getRegisterDefinitionAddress`

python/slsdet/detector.py
```diff
@@ -1813,6 +1814,109 @@ def reg(self):
...
+        Example
+        --------
+
+        d.define_reg('myreg',addr=0x6)
+        """
+        addr = RegisterAddress(addr)
+        self.setRegisterDefinition(name, addr)
+
+
+    def define_bit(self, name, addr, bit):
```
erikfrojdh: Should maybe also take str for the addr in case someone defined the register first.

python/slsdet/detector.py
```diff
@@ -1813,6 +1814,109 @@ def reg(self):
...
+        mybit = BitAddress(myreg, 5) 
+        d.getBit(mybit)
+        """
+        #Old usage passing two ints
+        if isinstance(bit_or_addr, int):
+            return super().getBit(bit_or_addr, number)
+
+        #New usage with str or BitAddress
+        if isinstance(bit_or_addr, str):
+            bit_or_addr = self.getBitDefinition(bit_or_addr)
```
erikfrojdh: Now needs `d.getBitDefinitionAddress` but why is address now added. This is really a bit definition

erikfrojdh (changes requested):
Python API needs updating, and we really need API tests for python =)

python/slsdet/detector.py
```diff
@@ -1813,6 +1814,109 @@ def reg(self):
...
+
+        d.define_reg('myreg',addr=0x6)
+        """
+        addr = RegisterAddress(addr)
+        self.setRegisterDefinition(name, addr)
+
+
+    def define_bit(self, name, addr, bit):
+        """
+        [Ctb] Define a name for a bit in a register to be used later with setBit/clearBit/getBit
```
AliceMazzoleni99, line 1837: lets make these keyword only

python/slsdet/utils.py
thattil: Removed

python/slsdet/registers.py
thattil: Done

python/slsdet/detector.py
thattil: Because you can also define with bit name and a bitaddress..

python/slsdet/detector.py
thattil: Yes, ive added it

AliceMazzoleni99 (approved):

PR #1315, opened 2025-10-14
used incorrect indexing

### LINKED ISSUE #1314: CTB-ADC plotting error in GUI
##### *Distribution: 
RHEL7

##### *Detector type: 
 ChipTestBoard

developer

##### Priority:
High

##### *Describe the bug
When all 32 ADCs are enabled, the plotting int the CTB GUI works fine. when only one ADC (tried with ADC7) is enabled for plotting, it doesn't work, i.e. it seems to plot a 'random' other ADC channel.
the problem refers to the "plot" checkbox only, not the "enable" checkbox

##### Expected behavior
obvious.

##### To Reproduce
Put some analog input (sine wave) at one ADC channel. might be tricky for a random board, can use Patrick's setup (please ask).
Start Gui, try plotting with only ADC7 enabled for plotting and than all of the ADCs

##### Screenshots
-

##### Additional context
Dhanya wrote it might have to do with the "reorder" that is now enable per default. At least Patrick, Martin M and Julian are not aware what this is and what it does.
We try to debug a bug in the firmware with this plotting, which is a bit tricky if the plotting is not reliable...
Patrick is willing to pay a coffee for whoever can solve this.

AliceMazzoleni99:
- indexing was probably wrong 
- reading the data for all enabled ADC but indexing only updated if plotting was enabled as well 
- generally plotted first n ADC's for which plotting was enabled as well

### DIFF
--- a/RELEASE.txt
+++ b/RELEASE.txt
@@ -23,11 +23,14 @@ This document describes the differences between vx.x.x and vx.0.2
 
 
 
-
 1 New, Changed or Resolved Features
 =====================================
 
+1.8 Gui
+----------
 
+Bug Fix: 
+    Plotting specific ADC as waveform plots correct data 
  
 
 2  On-board Detector Server Compatibility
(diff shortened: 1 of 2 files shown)


mmarti04 (approved):
Tested, works

thattil (commented):
The release notes will not have anything to do with the ctb or xilinx_ctb in there. We can remove this in the next PR.

PR #1316, opened 2025-10-15
Dev/xilinx set power

thattil:
- power set and retrieved similar to ctb, except without vchip.

Issues for both:
cannot set to -100 without issues. reg set only if dac>=0. Should also set if -100? Ignore -100 and this goes into another #1317  issue.


mmarti04 (approved):
tested, behaves now like the altera CTB

thattil:
> cannot set to -100 without issues. reg set only if dac>=0. Should also set if -100?
Alright, but both of them cannot really set to -100.. it will but complain and say it is 0. Do you want it to return -100 if that was set?

PR #1318, opened 2025-10-17
added option to use system zmq

### LINKED ISSUE #1308: Making building and linking libzmq statically configurable
##### *Detector type: 
Not Applicable

10.0.0

##### Priority:

Low

##### *State the change request:

`libzmq` is currently built from source and linked statically with the Python module, which is nice when building a wheel.

But that's an issue when packaging for other package managers like Debian or conda.
It should be possible to build the Python module against an existing `libzmq`.

##### Is your change request related to a problem. Please describe:

When building a conda package of `slsdet`, we should be able to depend on `libzmq` provided by the `zeromq` package.

erikfrojdh:
Added an option SLS_USE_SYSTEM_ZMQ to disable building libzmq as a part of the project
closes #1308

Should maybe be back ported to 10.1, maybe even 9.3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ include(cmake/SlsAddFlag.cmake)
 include(cmake/helpers.cmake)
 
 
+option(SLS_USE_SYSTEM_ZMQ "Use system installed libzmq" OFF)
 
 # Using FetchContent to get libzmq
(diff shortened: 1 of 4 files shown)


thattil (approved):
LGTM. If possible, I would also print the Zeromq version it found in the system.

PR #1321, opened 2025-10-21
Experimental support for using the client on macOS (Darwin)

erikfrojdh:
Experimental support for building the client (including python extension) on macOS. Enable controlling a detector from an Apple computer. Receiver still needs to run under linux. 

**Known issues:** 

- One test for SharedMemory disabled since macOS does not expose /dev/shm to the user
- One test for the udp receiving socket disabled since it it only used in the receiver 
- MacAddress from interface name disabled (throws) - look at it again if we want to use the receiver


slsSupportLib/src/file_utils.cpp
```diff
@@ -246,21 +251,39 @@ std::vector<int> getChannelsFromFile(const std::string &fname) {
...
+
+
+    #if defined(__APPLE__)
+    uint32_t size = PATH_MAX;
+    if (_NSGetExecutablePath(path.data(), &size) != 0) {
+        throw std::runtime_error("Failed to get executable path");
+    }
+    // Resolve any symlinks and .. components
+    char resolved[PATH_MAX];
+    if (!realpath(path.data(), resolved)) {
```
AliceMazzoleni99: Not sure I understand. Isnt it better to use string here as well. Same reasoning as above.

slsSupportLib/src/network_utils.cpp
```diff
@@ -16,7 +16,7 @@
 #include <netdb.h>
 #include <sstream>
 #include <sys/ioctl.h>
-#include <sys/prctl.h>
+// #include <sys/prctl.h>
 #include <sys/socket.h>
```
AliceMazzoleni99, line 19: Minor:remobe

slsSupportLib/src/network_utils.cpp
```diff
@@ -202,7 +208,8 @@ MacAddr InterfaceNameToMac(const std::string &inf) {
     if (sock != 1) {
         close(sock);
     }
-    return MacAddr(mac);
+    return MacAddr();
```
AliceMazzoleni99: why do we need mac if not passed to constructor

slsSupportLib/src/network_utils.cpp
erikfrojdh: Fixed

slsSupportLib/src/network_utils.cpp
erikfrojdh: Good catch! Copy paste error!

AliceMazzoleni99 (approved):
Oke looks good. Can we add the build on macos to github workflows or doesnt that work. I could also test it on my personal mac but I don't think it adds any value.

slsSupportLib/src/file_utils.cpp
erikfrojdh: I think using a string would be better. I user a char array since it was in the old implementation. Now updated and fixed missed assignment. Merging after tests pass.

PR #1324, opened 2025-10-23
Using find_package(Threads REQUIRED) instead of linking pthread directly

erikfrojdh:
- Using find_package(Threads) instead of directly linking to pthread to be more portable. 
- Still check that the found library is pthread compatible since we use direct calls to pthread internally. 
- Added Threads::Threads as a public dependency on slsDetectorSupport which means it will propagate to the other libraries and executables
- Removed redundant linking to rt
- Commented out failing test in gitea (revisit later)
- CI now includes virtual detector servers to catch build failures 

On linux link to rt even if we don't need it to avoid too many conditionals
 - If glibc<2.34 we need to link rt and both conda and wheels use older glibc

### DIFF
--- a/slsReceiverSoftware/CMakeLists.txt
+++ b/slsReceiverSoftware/CMakeLists.txt
@@ -118,7 +118,6 @@ if (SLS_USE_RECEIVER_BINARIES)
     target_link_libraries(slsReceiver PUBLIC
     PUBLIC
         slsReceiverStatic
-        rt
     PRIVATE
         slsProjectWarnings
     )
@@ -137,7 +136,6 @@ if (SLS_USE_RECEIVER_BINARIES)
     target_link_libraries(slsMultiReceiver 
     PUBLIC
         slsReceiverStatic
-        rt
     PRIVATE
         slsProjectWarnings
     )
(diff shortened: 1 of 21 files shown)


slsReceiverSoftware/CMakeLists.txt
AliceMazzoleni99, line 1: target_link_libraries (slsReceiverObject PUBLIC $<$<PLATFORM_ID:Linux>:rt>) dont we have to add this as well

AliceMazzoleni99: or maybe just link to slsSupportObject

AliceMazzoleni99 (commented):
Looks good. Added a few command where I was unsure.

slsReceiverSoftware/CMakeLists.txt
erikfrojdh: Linking it to slsSupportObject probably makes most sense since we have the shared memory implementation there and shm_open seems to be what the linker often complains about. In the receiver we don't use shm so there it is technically not needed but it doesn't hurt to link it. 

Once we move from libc 2.17 for the conda pkgs then we can drop rt completely since on modern systems most functions have been merged to libc

AliceMazzoleni99 (approved):

slsReceiverSoftware/CMakeLists.txt
thattil: shouldnt Release notes be updated?

PR #1327, opened 2025-10-28
Using system libzmq in conda builds

erikfrojdh:
- Using system libzmq by passing --config-settings=cmake.define.SLS_USE_SYSTEM_ZMQ=ON for conda builds
- wheels for pypi still uses the internal libzmq
- added python 3.14 to automatic builds
- bumped cibuildwheel version and therefore also the manylinux image to 2_28

### DIFF
--- a/.github/workflows/build_wheel.yml
+++ b/.github/workflows/build_wheel.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Build wheels
-        run: pipx run cibuildwheel==2.23.0
+        run: pipx run cibuildwheel==3.2.1
 
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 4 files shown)


AliceMazzoleni99 (approved):
Looks good. But why do we only use the system version for conda packages and not when using pip.

erikfrojdh:
> Looks good. But why do we only use the system version for conda packages and not when using pip.

With pip we have less control over which libraries are installed. We could try to find the .so from the pypi package of pyzmq but not sure how reliable that is. 

For conda it seems to be mandatory if we want slsDetectorPackage on conda-forge

thattil:
@erikfrojdh Should this go into release notes?

erikfrojdh:
I don't think that's necessary. It will have no observable effect...

PR #1330, opened 2025-11-14
1001/python cmd for roi

AliceMazzoleni99:
Previously ROI could not be set from python.
- roi can be set from Python 
- added tests 
- made "Parse port and uid" test hidden as it fails on gitea


.github/workflows/cmake.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
         with:
           python-version: 3.12
           cache: 'pip'
-      - run: pip install pytest numpy
+      - run: pip install pytest numpy colorama
```
thattil, line 22: i thought we didnt want to install it for the whole project?

RELEASE.md
```diff
@@ -0,0 +1,245 @@
+SLS Detector Package Minor Release 10.0.1 released on ?
+================================================================
+
```
thattil, line 3: Nice that you start with .md!

RELEASE.md
```diff
@@ -0,0 +1,245 @@
...
+
+1.2 New or Changed Features
+============================
+
+    Python
+    -------
+
+    * receiver ROI can be set from Python using command ``rx_roi``(it supports any sequence of four ints e.g. a tuple (xmin, xmax, ymin, ymax) or a sequence of such for multiple ROIS)
+    * one can clear all ROI's from Python using command ``clear_rx_roi``
```
thattil: I would use rx_clearroi because it would be consistent with the command line commands.

thattil (changes requested):
Didnt check the tests or the typecaster. sorry.

python/slsdet/detector.py
```diff
@@ -301,6 +302,40 @@ def rx_arping(self):
...
+        Each ROI is represented as a tuple of (x_start, y_start, x_end, y_end). \n
+        If no ROIs are configured, returns an empty list.
+        """
+        return self.getRxROI() #vector of Roi structs how represented? 
+    
+    @rx_roi.setter
+    def rx_roi(self, rois):
+        """Sets the list of ROIs in the receiver.
+
+        Note
```
thattil, line 324: Could you please also add the following info that is there in Detector.h as well?
/*only at multi module level without gap pixels. If more than 1 ROI per
     * UDP port, it will throw. Setting number of udp interfaces will clear the
     * roi. Cannot be set for CTB or Xilinx CTB */

python/slsdet/detector.py
```diff
@@ -301,6 +302,40 @@ def rx_arping(self):
...
+            raise TypeError(
+            "setRxROI failed: expected a tuple/list of ints x_min, x_max, y_min, y_max "
+            "or a sequence of such."
+        )
+        if(not isinstance(rois[0], Sequence)): 
+            self.setRxROI([rois])
+        else:
+            self.setRxROI(rois)
+
+    def clear_rx_roi(self):
```
thattil: Please change this to rx_clearroi to keep it consistent with command line cmds.

slsReceiverSoftware/tests/test-Apps.cpp
```diff
@@ -138,7 +138,8 @@ TEST_CASE("Parse version and help", "[detector]") {
     }
 }
 
-TEST_CASE("Parse port and uid", "[detector]") {
+// TODO: fails on gitea CI due to uid issue, fix later
```
thattil, line 141: Add this workaround for gitea fail in the description of the PR if you plan to add it with this PR

tests/scripts/utils_for_test.py
```diff
@@ -193,14 +193,28 @@ def connectToVirtualServers(name, num_mods, ctb_object=False):
...
+        cmd = ['slsReceiver']
+    else:
+        cmd = ['slsMultiReceiver', str(DEFAULT_TCP_RX_PORTNO), str(num_mods)]
+        # in 10.0.0
+        #cmd = ['slsMultiReceiver', '-p', str(DEFAULT_TCP_RX_PORTNO), '-n', str(num_mods)]
+    startProcessInBackground(cmd, fp)
+    time.sleep(1)
+
+
+def loadConfig(name, rx_hostname = 'localhost', settingsdir = None, log_file_fp = None, num_mods = 1, num_frames = 1, num_interfaces = 1):
```
thattil, line 206: Please fix these tests as discussed already. (test_roi.py, test_simulator.py, test_frame_syncrhonizer.py) and also "pytest -s test_free.py".. as they all depend on the utils_for_test.py

thattil (changes requested):

python/src/typecaster.h
```diff
@@ -97,5 +98,42 @@ template <> struct type_caster<std::chrono::nanoseconds> {
...
+
+        // accept tuple, list, numpy array any sequence
+        py::sequence seq;
+        try {
+            seq = py::reinterpret_borrow<py::sequence>(roi);
+        } catch (...) {
+            return false;
+        }
+
+        if (seq.size() != 4)
```
thattil: The sequence can also be a size of 2 right? for m3 and g2

tests/scripts/utils_for_test.py
AliceMazzoleni99: Done.

slsReceiverSoftware/tests/test-Apps.cpp
AliceMazzoleni99: done.

thattil (changes requested):

RELEASE.md
AliceMazzoleni99: done.

python/slsdet/detector.py
AliceMazzoleni99: done.

python/src/typecaster.h
AliceMazzoleni99: yes good point. Adapted.

.github/workflows/cmake.yaml
AliceMazzoleni99: Its only in a github workflow using DEBUG build so I dont see why it should be a problem. But I agree we dont want to clutter the environment.

AliceMazzoleni99: Oke I will merge now. We can discuss next week - If we want it or not than I can still remove. For the release it should not matter.

PR #1335, opened 2025-11-24
New this build bin path script

mbrueckner-psi:
The old this_build_bin_path script is terrible. I replaced it with a new one.


this_build_bin_path.sh
```diff
@@ -1,39 +1,15 @@
...
+# Since this script could be sourced, $0 is not sufficent, BASH_SOURCE[0] is necessary
+SCRIPT_LOCATION="$(realpath ${BASH_SOURCE[0]})"
+SCRIPT_LOCATION="$(dirname ${SCRIPT_LOCATION})"
+BUILDBIN_LOCATION="${SCRIPT_LOCATION}/build/bin"
+
+if [ ! -d "${BUILDBIN_LOCATION}" ]; then
+	echo Cannot find path ${BUILDBIN_LOCATION}
+else
+	echo Adding ${BUILDBIN_LOCATION} to PATH, PYTHONPATH and LD_LIBRARY_PATH
+	export PATH=${PATH}:${BUILDBIN_LOCATION}
```
erikfrojdh: Do you want to put BUILDBIN first, in case there is another slsDetectorPackage installed?

this_build_bin_path.sh
```diff
@@ -1,39 +1,15 @@
...
+	export PYTHONPATH=${PYTHONPATH}:${BUILDBIN_LOCATION}
+	export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:${BUILDBIN_LOCATION}
```
erikfrojdh: I don't think we need LD_LIBRARY_PATH anymore since we link statically.

erikfrojdh (changes requested):

this_build_bin_path.sh
mbrueckner-psi: I prefer extending variables like PATH and LD_LIBRARY_PATH at the end to not break functionality of other programs.

mbrueckner-psi: but in this specific situation it might be useful. I agree

erikfrojdh (approved):

PR #1337, opened 2025-11-25
Dev/add simulator tests in GitHub workflows

AliceMazzoleni99:
- added simulator tests to github workflows
- a couple of tests fail on the runner as well as locally for me e.g. tests which compare number of frames (they fail frequently for jungfrau and xilinx_ctb) , v_abcd fails for xilinx


docs/src/Testing.rst
```diff
@@ -0,0 +1,105 @@
...
+
+If a test fails in the github or gitea actions hide the test by adding the tag ``[.]`` to the test name.
+
+.. code-block:: cpp
+
+    TEST_CASE("This test is hidden from default runs", "[.]") {
+        REQUIRE(1 == 2);
+    }
+
+If a test requires virtual detector mark them with the hidden tag ``[.cmdcall]``.
```
thattil: .cmdcall is for any integration test involving a detector (be it virtual or real)..

docs/src/Testing.rst
```diff
@@ -0,0 +1,105 @@
...
+If you want to disable tests for a specific detector add the hidden tag ``[.disable_<detector_name>]`` to the test case.
+
+If you want to run tests requiring detector simulators run them as follows: 
+
+.. code-block:: console
+
+    cd build 
+    python bin/test_simulators.py 
+
+This runs all tests marked with the tag ``[.cmdcall]`` for all detector simulators.
```
thattil: Maybe you could add the following so that they are aware of this option. "To run all tests including the other tests (general tests not requiring a detector server), use -g or --general-tests option" or something like that.

docs/src/Testing.rst
```diff
@@ -0,0 +1,105 @@
...
+If you want to run them for a specific virtual detector or a specific test use the following command:
+
+.. code-block:: console
+    cd build
+    python bin/test_simulators.py --servers <detector_name> --test <test_name>
+
+You can run all tests requiring simulators but the disabled one for a specific detector e.g. marked with tag ``[.disable_<detector_name>]`` by adding the option ``--disable_<detector_name>``
+
```
thattil, line 74: the disable detector is a nice feature!

docs/src/Testing.rst
```diff
@@ -0,0 +1,105 @@
...
+.. code-block:: console
+    cd build
+    python bin/test_simulators.py --servers <detector_name> --test <test_name> --disable_jungfrau
+
```
thattil, line 78: I assume here one should remove the "--servers"? I assume servers and _disable are mutually exclusive options?

docs/src/Testing.rst
```diff
@@ -0,0 +1,105 @@
...
+To run the python tests use the following commands:
+
+.. code-block:: console
+
+    cd build
+    cmake ../ -DSLS_USE_PYTHON=ON
+    export PYTHONPATH=$PWD/bin 
+    python -m pytest ../python/tests/ 
+
+If a test requires a virtual detector mark them with the pytest marker ``@pytest.mark.withdetectorsimulators``.
```
thattil: " a detector server" should do I assume? I assume this works for both virtual and real detectors?

docs/src/Testing.rst
```diff
@@ -0,0 +1,105 @@
+Testing
+==========
```
thattil, line 2: Thanks for adding this to the documentation. It is done well too :)

docs/src/index.rst
```diff
@@ -66,6 +66,7 @@ slsDetectorPackage
     container_utils
     type_traits
     ToString
+    Testing
```
thattil, line 69: Perfect place to add it

docs/src/Testing.rst
thattil: also, the .cmdcall is by default included in the simulators python script.. if one needs a different market, one has to explcitly set the marker to that..

slsDetectorSoftware/tests/Caller/test-Caller-acquire.cpp
```diff
@@ -16,8 +16,11 @@ namespace sls {
 using test::GET;
 using test::PUT;
 
+// disable for jungfrau as it requires higher maximum receive buffer size
+//  sysctl net.core.rmem_max=$((100*1024*1024))
+//  sysctl net.core.rmem_default=$((100*1024*1024))
 TEST_CASE("jungfrau_or_moench_acquire_check_file_size",
-          "[.cmdcall][.cmdacquire]") {
+          "[.cmdcall][.cmdacquire][.disable_jungfrau]") {
```
thattil, line 24: is this disabled for Jungfrau for all tests?

slsDetectorSoftware/tests/Caller/test-Caller-acquire.cpp
```diff
@@ -150,7 +153,11 @@ void test_ctb_file_size_with_acquire(Detector &det, Caller &caller,
                                                   expected_image_size));
 }
 
-TEST_CASE("ctb_acquire_check_file_size", "[.cmdcall][.cmdacquire]") {
+// disable for jungfrau as it requires higher maximum receive buffer size
+//  sysctl net.core.rmem_max=$((100*1024*1024))
```
thattil, line 160: this seems to be with xilinx_ctb instead of jungfrau

slsDetectorSoftware/tests/Caller/test-Caller-chiptestboard.cpp
```diff
@@ -1056,9 +1056,10 @@ TEST_CASE("v_abcd", "[.cmdcall]") {
                 REQUIRE(oss2.str() == cmds[i] + " 1200\n");
             }
             for (int i = 0; i != det.size(); ++i) {
+                // TODO: indices should be independant of detector size? is i
+                // the correct index?
```
thattil: Good catch! It was bad variable naming. 'i' is reused as a localvariable when its also used for loop over the power indices.

slsDetectorSoftware/tests/Caller/test-Caller-global.cpp
```diff
@@ -172,9 +172,13 @@ void create_files_for_acquire(
 
     // acquire and get num frames caught
     REQUIRE_NOTHROW(test_acquire_with_receiver(caller, det));
-    auto frames_caught = det.getFramesCaught().tsquash(
-        "Inconsistent number of frames caught")[0];
-    REQUIRE(frames_caught == num_frames);
+    // TODO: maybe there should not be REQUIRE statements in void function at
+    // all
```
thattil: What would be the reason why this should not be allowed? Maybe I am missing something.

slsDetectorSoftware/tests/Caller/test-Caller-master-attributes.cpp
```diff
@@ -1046,7 +1046,7 @@ TEST_CASE("check_master_file_attributes", "[.cmdcall][.cmdacquire][.cmdattr]") {
     case defs::MOENCH:
     case defs::MYTHEN3:
     case defs::GOTTHARD2:
-        create_files_for_acquire(det, caller, num_frames);
+        create_files_for_acquire(det, caller, num_frames, std::nullopt, false);
         break;
```
thattil, line 1052: the check is hard coded to a 'false' so the frames caught will never happen for these tests. Something to check if this is really needed.

slsDetectorSoftware/tests/Caller/test-Caller-rx-running.cpp
```diff
@@ -141,13 +141,15 @@ TEST_CASE("numinterfaces - cant put if receiver is not idle",
...
+
+        auto numinterfaces = det.getNumberofUDPInterfaces();
 
         // stop receiver
         REQUIRE_NOTHROW(caller.call("rx_stop", {}, -1, PUT));
 
         for (int i = 0; i != det.size(); ++i) {
-            det.setNumberofUDPInterfaces(prev_numinterfaces[i], {i});
+            det.setNumberofUDPInterfaces(prev_numinterfaces[i]);
         }
```
thattil: why is the position index ignored here? if we are going to set this "previous value" for all the modules, we should then also specify which module. Ideally, this should be consistent across modules, but then we should also not have an array and instead just have one value for all.

slsDetectorSoftware/tests/Caller/test-Caller-rx.cpp
```diff
@@ -756,7 +756,8 @@ TEST_CASE("rx_roi", "[.cmdcall]") {
         // check master file creation
         // TODO: check roi in master file
         {
-            REQUIRE_NOTHROW(create_files_for_acquire(det, caller));
+            REQUIRE_NOTHROW(
+                create_files_for_acquire(det, caller, 1, std::nullopt, false));
             testFileInfo file_info;
```
thattil, line 760: Again, hard coded false to not test. To look into.

slsReceiverSoftware/src/Implementation.cpp
```diff
@@ -813,8 +813,9 @@ void Implementation::startReadout() {
                     << "waiting for all packets, previousValue:"
                     << previousValue
                     << " totalPacketsReceived: " << totalPacketsReceived;
-                /* TODO! Need to find optimal time **/
-                std::this_thread::sleep_for(std::chrono::milliseconds(5));
+                /* TODO! Need to find optimal time - xilinx and jungfrau need
+                 * more time */
+                std::this_thread::sleep_for(std::chrono::milliseconds(20));
                 previousValue = totalPacketsReceived;
```
thattil, line 819: That is strange as it worked before :) I will look into this.

slsReceiverSoftware/tests/test-ArrangeDataBasedOnBitList.cpp
```diff
@@ -168,7 +168,7 @@ TEST_CASE_METHOD(DataProcessorTestFixture, "Remove Trailing Bits",
 // parametric test tested with num_samples = 5, num_samples = 10, num_samples =
 // 8
 TEST_CASE_METHOD(DataProcessorTestFixture, "Reorder all",
-                 "[.dataprocessor][.reorder]") {
+                 "[dataprocessor][reorder]") {
```
thattil, line 173: Thank you! missed the '.'..

tests/CMakeLists.txt
```diff
@@ -60,5 +60,3 @@ catch_discover_tests(tests)
 configure_file(scripts/test_simulators.py ${CMAKE_BINARY_DIR}/bin/test_simulators.py COPYONLY)
 configure_file(scripts/test_frame_synchronizer.py ${CMAKE_BINARY_DIR}/bin/test_frame_synchronizer.py COPYONLY)
 configure_file(scripts/utils_for_test.py ${CMAKE_BINARY_DIR}/bin/utils_for_test.py COPYONLY)
```
thattil, line 62: I see that test_roi.py has been removed from this folder but I dont see it added anywhere else. Is there a reason for this?

tests/scripts/test_simulators.py
```diff
@@ -2,91 +2,114 @@
...
+        test_filter += " ~[disable_gotthard2]"
+    if args.disable_eiger: 
+        test_filter += " ~[disable_eiger]"
+
+    cmd = [str(build_dir / 'tests'), '--abort', test_filter, '-s'] 
 
-def startCmdTestsForAll(args, fp):
     for server in args.servers:
-        try:
-            num_mods = 2 if server == 'eiger' else 1
```
thattil, line 58: Is  this eiger default 'num mod = 2' set somewhere?

tests/scripts/utils_for_test.py
```diff
@@ -289,16 +343,29 @@ def ParseArguments(description, default_num_mods=1, markers=False, general_tests
...
+    parser.add_argument("-d_jungfrau", "--disable_jungfrau", action="store_true", help="Disable all tests with hidden tag [.disable_jungfrau]")
+    parser.add_argument("-d_eiger", "--disable_eiger", action="store_true", help="Disable all tests with hidden tag [.disable_eiger]")
+    parser.add_argument("-d_moench", "--disable_moench", action="store_true", help="Disable all tests with hidden tag [.disable_moench]")
+    parser.add_argument("-d_gotthard2", "--disable_gotthard2", action="store_true", help="Disable all tests with hidden tag [.disable_gotthard2]")
+    parser.add_argument("-d_mythen3", "--disable_mythen3", action="store_true", help="Disable all tests with hidden tag [.disable_mythen3]")
+    parser.add_argument("-d_ctb", "--disable_ctb", action="store_true", help="Disable all tests with hidden tag [.disable_ctb]")
 
+    
+    if specific_tests:
+        parser.add_argument('-t', '--tests', nargs='?', default ='[.cmdcall]',
```
thattil: Much more readable name than 'markers'! Nice

tests/scripts/utils_for_test.py
```diff
@@ -289,16 +343,29 @@ def ParseArguments(description, default_num_mods=1, markers=False, general_tests
                         help='Number of frames to test with')
     parser.add_argument('-s', '--servers', nargs='*',
                         help='Detector servers to run')
-    if markers:
-        parser.add_argument('-m', '--markers', nargs='?', default ='[.cmdcall]',
-                        help = 'Markers to use for cmd tests, default: [.cmdcall]')
+    parser.add_argument('-nlf', '--no-log-file', action='store_true',
```
thattil, line 373: Nice add on

thattil (changes requested):
Over all, its good. I still have to go through a few things. I will try to figure out this week why it is failing, and what these changes mean.

docs/src/Testing.rst
AliceMazzoleni99: If you run python bin/test_simulators.py --servers jungfrau --test [.cmdcall] --disable_jungfrau. It will still run all tests with [.cmdcall] for jungfrau expect the ones which are marked with [.disable_jungfrau]. I added these options as for example the  tests where I need to set the kernel parameters will fail on the girhub runner.

docs/src/Testing.rst
AliceMazzoleni99: I will change the name to [.detectorintegration] as discussed I think it makes more sense.

docs/src/Testing.rst
```diff
@@ -0,0 +1,105 @@
...
+.. code-block:: cpp
+
+    TEST_CASE("This test is hidden from default runs", "[.]") {
+        REQUIRE(1 == 2);
+    }
+
+If a test requires virtual detector mark them with the hidden tag ``[.cmdcall]``.
+
+If you want to disable tests for a specific detector add the hidden tag ``[.disable_<detector_name>]`` to the test case.
+
```
AliceMazzoleni99, line 53: I will add here If you want to disable a specific test for a specific detector... . I think it makes it clearer.

docs/src/Testing.rst
AliceMazzoleni99: Jep will change. Will also change the market to pytest.mark.detectorintegration

slsDetectorSoftware/tests/Caller/test-Caller-acquire.cpp
AliceMazzoleni99: ah copy paste error will update.

slsDetectorSoftware/tests/Caller/test-Caller-global.cpp
AliceMazzoleni99: I was thinking that it should just be a helper function and the actual checks should be done in the test itself. But as its a Require i guess its okay. Its also easier for the clean up.

slsReceiverSoftware/src/Implementation.cpp
AliceMazzoleni99: I will change it back. Was from debugging. I needed to reset the receiver buffer size.

tests/CMakeLists.txt
AliceMazzoleni99: All the roi tests run in test_simulators.py

tests/scripts/test_simulators.py
AliceMazzoleni99: Nice catch need to update.

docs/src/Testing.rst
AliceMazzoleni99: added it.

slsDetectorSoftware/tests/Caller/test-Caller-acquire.cpp
AliceMazzoleni99: I dont get it? Only for this one test case. But even if you specify one hidden tag it will run it.

slsDetectorSoftware/tests/Caller/test-Caller-rx-running.cpp
AliceMazzoleni99: The test fails if I set it one by one for each module. getValidPortNumbers throws an error if modules have different number of udp ports. But you're right its ugly. Cleaned up the test squashed getNumberofUDPInterfaces etc.

slsDetectorSoftware/tests/Caller/test-Caller-rx.cpp
AliceMazzoleni99: It repeatedly fails for jungfrau if one does not update the maximum buffer size - which I cant on github runners. As for this test the number of frames is not what we want to test anyway I dont check for it - to make things simpler.

tests/scripts/test_simulators.py
AliceMazzoleni99: done.

slsDetectorSoftware/tests/Caller/test-Caller-master-attributes.cpp
AliceMazzoleni99: Changed it such that its only set to false for jungfrau.

docs/src/Testing.rst
thattil: Suggestion to be more clear? => "You can run all tests requiring a detector except the disabled ones.."

tests/scripts/test_simulators.py
```diff
@@ -2,91 +2,121 @@
 # Copyright (C) 2021 Contributors to the SLS Detector Package
 '''
 This file is used to start up simulators, receivers and run all the tests on them and finally kill the simulators and receivers.
+
+It can be used to run all catch tests with tag [.detectorintegration]. 
+
+Pass --tests <testname> to run specific tests only or --tesst <testtag> to run all tests with that specific tag.
```
thattil: tests typo

docs/src/Testing.rst
thattil: eg. python bin/test_simulators.py --servers jungfrau eiger ctb --test "[.cmdcall] ~[.disable_jungfrau]"

PR #1338, opened 2025-11-26
Local runners & NFS deploy

mmarti04:
Workflow running on our own machines (registered in gitea), both in the server room so they should not move to beamlines etc. by accident.

everything in gitea_runner@mpc2935:/slsDetectorSoftware/ is automatically deployed to NFS to /psi.ch/group/detector/software/slsDetectorPackage/developer each morning at 5

### DIFF
--- a/.gitea/workflows/rh8-local.yml
+++ b/.gitea/workflows/rh8-local.yml
@@ -1,0 +1,27 @@
+name: Build on local RHEL8
+
+on:
+  push:
+    branches:
+      - dev/NFS_testing
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: "detectors-software-RH8"
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build library 
+        run: |
+            mkdir build && cd build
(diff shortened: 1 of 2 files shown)


.gitea/workflows/rh8-local.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  push:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: "detectors-software-RH8"
```
AliceMazzoleni99, line 14: How did you do that? Is it a self hosted runner?

.gitea/workflows/rh8-local.yml
```diff
@@ -0,0 +1,25 @@
+name: Build on local RHEL8
+
+on:
+  push:
+  workflow_dispatch:
+
```
AliceMazzoleni99, line 8: do we want to run this on every push or only for push into developer? Its deployed only for pushes/merges into developer

.gitea/workflows/rh8-local.yml
mmarti04: yes, in gitea

AliceMazzoleni99 (commented):
Maybe only build on local pc if pushed into developer. Otherwise nice thing with the custom runners. I didnt know one can do that.

AliceMazzoleni99 (approved):

PR #1344, opened 2025-12-11
Using define reg/bit from python

erikfrojdh:
Missing: 

- [ ] tests
- [x] clearbit

Usage
```python
# --- mh02.defs ---
# define addr some_reg 0x5
# define addr another_reg 0x6
# define bit mybit some_reg 7

# --- test.config ---
# hostname localhost
# parameters mh02.defs


d = Detector()
#g reg 0x5
d.reg[0x5] 

#g reg somereg
d.reg['somereg']

d.reg['anotherreg'] = 78
d.reg[0x6] = 11

d.define_reg('yetanotherreg', 0x7)
d.define_reg('evenmoreregs', RegisterAddress(0x8))
d.define_reg('randomreg', addr = 0x9)

d.define_bit('mybit2', 0x1, 3)
# d.setBit('mybit2')

d.setBit(0x5,3)
d.setBit('mybit2')
bit4 = BitAddress(RegisterAddress(0x7),4)
d.setBit(bit4)

```


python/slsdet/registers.py
```diff
@@ -1,13 +1,28 @@
...
 # Copyright (C) 2021 Contributors to the SLS Detector Package
+
+from ._slsdet import RegisterValue, RegisterAddress
+from .utils import element
+
 class Register:
     def __init__(self, detector):
         self._detector = detector
 
+    @element
```
AliceMazzoleni99, line 11: is an element neccessary here? the names should be unique

erikfrojdh: The detector API will return a vector of length num modules. In the case of ctb length one. Using the element decorator saves the user from having to select the 0th element (res = reg[addr][0])

PR #1347, opened 2025-12-16
clang-tidy [unhandled-self-assignment]

erikfrojdh:
Handle self assignment in sls_detector_module. 

https://clang.llvm.org/extra/clang-tidy/checks/bugprone/unhandled-self-assignment.html

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -29,7 +29,6 @@ Checks:          '*,
                     -llvmlibc-*'
 
 HeaderFilterRegex: \.h
-AnalyzeTemporaryDtors: false
 FormatStyle:     none
 CheckOptions:
   - { key: readability-identifier-naming.NamespaceCase,       value: lower_case }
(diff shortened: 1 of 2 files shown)


.clang-tidy
AliceMazzoleni99, line 1: shouldnt one add -bugprone-unhandled-self-assignment in the Checks or do I miss something?

AliceMazzoleni99 (commented):
Nice Catch. Should we consider using git pre commit hooks or run it during build or github workflows to run clang-tidy?

.clang-tidy
erikfrojdh: At the moment we have 
`
Checks:          '*, 
`
In the clang-tidy file so it is already enabled. We can discuss long term if we want to enable all and then disable some, like now or only enable the ones we want.

erikfrojdh:
> Nice Catch. Should we consider using git pre commit hooks or run it during build or github workflows to run clang-tidy?

At the moment this is not possible since we have too many errors. We should work through them first and then enable this once we have a clean build.

.clang-tidy
AliceMazzoleni99: ah missed that

AliceMazzoleni99 (approved):

PR #1348, opened 2025-12-16
marked unused functions readDataFile/writeDataFile deprecated

erikfrojdh:
readDataFile and writeDataFile in file_utils.h doesn't seem to be used and I can't think of how external people would use them. 

Marked deprecated for now, but maybe we can remove them?


thattil:
I would say remove?

erikfrojdh:
Opened new PR since the branch is outdated. Look at #1456

PR #1350, opened 2025-12-18
fixed python simulator test fixture

AliceMazzoleni99:
- fixed test fixture for simulator test 
- test fixture has module scope (setup and teardown only happens once)
- yield only once per detector 
- second parametrized test fixture to allow to test for specific detectors


python/tests/test_ROI.py
```diff
@@ -40,32 +41,33 @@ def test_rx_ROI(test_with_simulators):
 
         assert d.rx_roi == [(5,15,15,25)]
 
-        if d.type != detectorType.JUNGFRAU and d.numinterfaces != 2:
+        if d.nmod > 1 and (d.type != detectorType.JUNGFRAU) or (d.numinterfaces == 2 and d.type != detectorType.EIGER):
```
thattil, line 42: what is this condition for?  The second condition i realize its for Jungfrau and moench 2 interfaces, but the first part? why only jungfrau? what is this condition checking?

tests/scripts/utils_for_test.py
```diff
@@ -196,62 +270,66 @@ def connectToVirtualServers(name, num_mods, ctb_object=False):
...
+    if no_log_file: 
+        fname = None
+    startProcessInBackground(cmd, fp, fname, quiet_mode)
+    time.sleep(1)
 
 def loadConfig(name, rx_hostname = 'localhost', settingsdir = None, log_file_fp = None, num_mods = 1, num_frames = 1, num_interfaces = 1):
-    Log(LogLevel.INFO, 'Loading config')
-    Log(LogLevel.INFO, 'Loading config', log_file_fp)
+    Log(LogLevel.INFO, 'Loading config', log_file_fp, True)
+    Log(LogLevel.INFORED, f"Config for {name}, num_mods={num_mods}, num_interfaces={num_interfaces}", log_file_fp, True)
```
thattil: Was this for debugging? Remove?

python/tests/test_free.py
```diff
@@ -7,47 +7,28 @@
...
+    SERVER_START_PORTNO
+)
 
+from conftest import session_simulator
 
-def test_exptime_after_free_should_raise(setup_simulator):
+@pytest.mark.detectorintegration
+@pytest.mark.parametrize("session_simulator",[("ctb", 1, 1)],indirect=True)
+def test_exptime_after_free_should_raise(session_simulator):
     Log(LogLevel.INFOBLUE, f'\nRunning test_exptime_after_free_should_raise')
```
thattil, line 33: As discussed now, loading config and basic settigns after free to pass the tests

python/tests/conftest.py
```diff
@@ -20,94 +18,113 @@
...
+def test_define_reg(session_simulator):
+    det_type, num_interfaces, num_mods, d = session_simulator
+'''
+@pytest.fixture(scope="session")
+def session_simulator(request):
+    """
+    Fixture to start the detector server once and clean up at the end.
+    Expects request.param = (det_type, num_interfaces, num_mods)
+    """
+    try:
```
thattil, line 81: 👍

thattil (changes requested):
Looks good overall, but tests failing because of test_free.py

thattil (approved):
The tests had been ongoing for the latest commit and the tests had failed only for the previous commit. My mistake.

I see that everything had passed. Also, regarding test_free.py, the tests are minor that config file and basic settings are not needed. This looks good. Thanks!

python/tests/test_free.py
thattil: Ignore this.

python/tests/test_ROI.py
AliceMazzoleni99: Jungfrau module is large enough such that the ROI would fit into one module which is not allowed - if they are not consecutive.

tests/scripts/utils_for_test.py
AliceMazzoleni99: Removed it.

tests/scripts/utils_for_test.py
```diff
@@ -287,9 +287,11 @@ def startReceiver(num_mods, fp, no_log_file = False, quiet_mode=False):
 
 def loadConfig(name, rx_hostname = 'localhost', settingsdir = None, log_file_fp = None, num_mods = 1, num_frames = 1, num_interfaces = 1):
     Log(LogLevel.INFO, 'Loading config', log_file_fp, True)
-    Log(LogLevel.INFORED, f"Config for {name}, num_mods={num_mods}, num_interfaces={num_interfaces}", log_file_fp, True)
     try:
-        d = connectToVirtualServers(name, num_mods)
+        if name == 'ctb' or name == 'xilinx_ctb':
+            d = connectToVirtualServers(name, num_mods, ctb_object=True)
```
thattil, line 292: 👍

thattil (approved):

PR #1353, opened 2026-01-07
Dev/add simulator tests in GitHub workflows fix

thattil:
workflows
- added gitea workflow for rh8 and rh9 runner, which tests with log files and running tests that check frames caught
- modified github workflow
   - run on all dev/* branches
   - run also hdf5 tests
   - also run python tests not including detector-integration marker
   - run simulator disabling tests with 'disable_check_data_file' instead of 'disable_jungfrau' and 'disable_xilinx_ctb'
- removed documentation about 'disable_jungfrau' and added about disabling tests.
- removed integration tests folder and configuration in CMakeLists.txt
- test_simulator runs now for 2 modules for all (except ctb and xilinx_ctb)
- improved python tests fixture to run for all modules, interfaces and detector types (the combos that work)


.gitea/workflows/rh8-local-tests.yml
```diff
@@ -0,0 +1,61 @@
+name: Run Simulator Tests on local RHEL8
+
+on:
+  push:
```
AliceMazzoleni99, line 4: should we just do it on pull-requests?

.github/workflows/run_tests.yaml
```diff
@@ -32,28 +29,27 @@ jobs:
...
+        run: |
+          python -m pytest -m detectorintegration ${{github.workspace}}/python/tests --detector-integration 
+          python -m pytest ${{github.workspace}}/python/tests
 
       - name: Simulator unit tests
         working-directory: ${{github.workspace}}/build/bin
         run: |
-            python test_simulators.py -nlf --disable_jungfrau --disable_xilinx_ctb
+            python test_simulators.py -nlf --tests '[detectorintegration] ~[disable_jungfrau] [disable_check_data_file]'
             python test_frame_synchronizer.py -nlf
```
AliceMazzoleni99, line 54: is ~disable_jungfrau still supported?

.github/workflows/run_tests.yaml
```diff
@@ -32,28 +29,27 @@ jobs:
...
       - name: Build
           # Build your program with the given configuration
         run: cmake --build ${{github.workspace}}/build -j4 --config ${{env.BUILD_TYPE}}
 
       - name: Python unit tests with simulators
         working-directory: ${{github.workspace}}/build/bin
-        run: python -m pytest -m detectorintegration ${{github.workspace}}/python/tests --detector-integration 
+        run: |
+          python -m pytest -m detectorintegration ${{github.workspace}}/python/tests --detector-integration 
+          python -m pytest ${{github.workspace}}/python/tests
```
AliceMazzoleni99, line 53: this we actually run already in cmake.yml but maybe we can remove there and only run here. Also add the general c++ tests here we anyway build twice

docs/src/Testing.rst
```diff
@@ -29,11 +29,23 @@ If a test fails in the github or gitea actions hide the test by adding the tag `
         REQUIRE(1 == 2);
     }
 
-If a test requires a detector mark them with the hidden tag ``[.detectorintegration]``.
+If a test requires a detector mark them with the hidden tag ``[.detectorintegration]``. This only works if a configured simulator (or an actual configured detector) and receiver are already set up to run the tests.
```
AliceMazzoleni99, line 33: Maybe better: You can run all tests requiring a detector by running the following command: 
.. code-block:: 

Note: This only works if a configured simulator (or an actual configured detector) and receiver are already set up to run the tests.

AliceMazzoleni99: There is a script that automatically sets up virtual detectors and runs all integration tests. See Section Simulator Script below.

python/tests/conftest.py
```diff
@@ -38,6 +38,93 @@ def pytest_collection_modifyitems(config, items):
...
+        num_frames=1, 
+        num_interfaces=num_interfaces,
+    )
+
+    loadBasicSettings(name=det_type, d=d, fp=fp)
+
+    yield det_type, num_interfaces, num_mods, d
+
+    cleanup(fp)
+
```
AliceMazzoleni99, line 92: this is really nice also that one can easily parametrize and it takes the same fixture. I had something similar in the fix/pytest_simulators PR just with module scope and another fixture that one can parametrize. But this is better. I think we can remove all other test fixtures.

.gitea/workflows/rh8-local-tests.yml
thattil: we discussed and decided to keep push just to have quick feedback on whichever hotfix or feature branch we are currently working on, instead of all of it at once when we do a PR.

.github/workflows/run_tests.yaml
thattil: fixed. Nice catch!

PR #1357, opened 2026-01-13
Dev/automate documentation build

### LINKED ISSUE #1356: Automate building documentation and upload for developer
As the documentation for developer changes frequently build and upload for every Pull Request.

AliceMazzoleni99:
- replaces devdoc with gh-pages 
- created GithUb workflow for automation 
- automatically adds new entry for a release to main index of versioned documentation in gh-pages
- automatically moves  documentation to gh-pages
- automatically moves RELEASE notes to documentation (if release) 
- automatically builds documentation 

TODO: 
- redirect files in devdoc to github.io/slsDetectorPackage - done
- maybe cache build in workflows ? - Done twice


.github/workflows/build_documentation.yml
```diff
@@ -0,0 +1,128 @@
...
+                path: gh-pages
+                
+            
+            - name: Copy documentation and Release notes to versioned folder
+              if: github.event_name == 'release' || (github.event_name == 'pull_request' && github.event.pull_request.merged == true)
+              run: |
+                VERSION="${{ steps.version.outputs.version }}"
+                mkdir -p "gh-pages/${VERSION}"
+                cp -r build/docs/html/. "gh-pages/${VERSION}/"
+                cp docs/main_index/index.html "gh-pages/index.html"
```
thattil: We dont really need the quotes for the destinations of cp in the second and third line

README.md
```diff
@@ -1,307 +0,0 @@
-## Dependencies
```
thattil, line 1: This shouldn't be deleted. 

Maybe only the basic Dependencies, Installation, Firmware and documentation links should be shown here.  The rest are just redundant

.github/workflows/build_documentation.yml
```diff
@@ -0,0 +1,128 @@
...
+            # Update main index if this is a release
+            - name: Update version index
+              if: github.event_name == 'release'
+              run: |
+                python docs/main_index/render_main_index.py \
+                --version "${{ steps.version.outputs.version }}" \
+                --date "$(date +'%d.%m.%Y')" 
+
+            # update paths to documentation version etc in release notes 
+            - name: Update Release Notes
```
thattil: I would not touch the release notes after release. I would remove "Update release notes" from the github workflow.

I would instead combine this script with the update_version.py. So when one runs this script, it will also update the release notes with the right version.

It will also not need to get the version from a command line argument. It should just pick it up from the VERSION file or with the version written to file.

etc/generate_release_notes.py
```diff
@@ -0,0 +1,61 @@
+from pathlib import Path
+from datetime import datetime
```
thattil: As mentioned earlier, I would remove this script and combine the essence of overwriting the RELEASE.md file with the correct versions (keeping the 'change log' intact) with the "update_version.py" script.

etc/generate_release_notes.py
```diff
@@ -0,0 +1,61 @@
...
+
+    prev_version = data['versions'][0]["version"]
+    if prev_version == version: #already updated to new version 
+        prev_version = data['versions'][1]["version"]
+
+    return prev_version
+
+def main(): 
+    parser = argparse.ArgumentParser(description='Update release notes from templated md file')
+    parser.add_argument('--version', type=str, help='Version to use in release notes, if not provided will read from VERSION file')
```
thattil: I dont think this argument would be needed in the future as one can use the version from the file or being written to file.

etc/generate_release_notes.py
```diff
@@ -0,0 +1,61 @@
...
+        version = args.version
+    else:
+        version = Path(ROOT_DIR / "VERSION").read_text().strip()
+
+    release_type = extract_release_type(version)
+
+    prev_version = get_previous_version(ROOT_DIR / "docs/main_index/versions.yaml", version)
+
+    # Read template
+    template = Path(ROOT_DIR / "RELEASE.md.template").read_text()
```
thattil: Read the same file 'RELEASE.md' instead of RELEASE,md.template to overwrite it.

RELEASE.md.template
```diff
@@ -45,11 +45,11 @@ added patternstart to python (ctb, xilinx_ctb , mythen3), only the detector clas
 ==========================================
 
 
-    Eiger       10.0.0
-    Jungfrau    10.0.0
-    Mythen3     10.0.0
-    Gotthard2   10.0.0
-    Moench      10.0.0
+    Eiger       {{VERSION}}
```
thattil: Only for major versions, these server versions get replaced automatically. So, I would not have it like this for every version. I would call it a different variable and update it whenever there is a major version only. For other versions, one has to update it manually as it is updated only via PRs  (minor features or bug fixes). 

For major versions, one can also do this manually instead of automatically. So upto you for major versions.

thattil (changes requested):
Looks awesome. Just a few changes such as not touching release notes when releasing and when the user checks out the branch, the release notes are final. Otherwise, it looks great!

.github/workflows/build_documentation.yml
AliceMazzoleni99: removed them

README.md
AliceMazzoleni99: deleted by mistake. Updated paths in README.

.github/workflows/build_documentation.yml
AliceMazzoleni99: done.

RELEASE.md.template
AliceMazzoleni99: left old versions. One has to manually change it upon a release

PR #1358, opened 2026-01-14
Dev/xilinx and ctb setpower and setdac in server

thattil:
Xilinx
- power regulators do not set power down value anymore (setting 1 to power down file instead sets max voltage due to schematics),Therefore, the default is to set them to their minimum value with power disabled.
- separate setDAC and setPower. 
   - SetPower checks its own indices, does its own mV to dac value conversion, checks its own limits including vlimit, ensures no power down value (-100) given to it and then calls LTC2620 WriteDac directly.
   - SetDAC checks its own indices, vlimits and then calls LTC2620 SetDac to convert to dac values (if given in mV), also print in mV (if given in dac values) and then calls LTC2620 WriteDAC directly.
   - They both now return ret and is handled in funcs.c.
- fixed transceiver alignment for virtual xilinx servers 

LTC2620 (affected by xilinx, m3 and g2)
- LTC2620 SetDAC has been refactored into SetDac and WriteDac for readability. SetDac only called by Dacs (not power regulators) as mentioned before, but writeDac writes to file.
- removed "%d" from the macros as in the end they are used in variables and this is 'unsafe'. Power down file name is only for xilinx and needs to be concatenated and therefore removed from LTC2620 variable name


mmarti04 (changes requested):
power to 0 is still a self-destruct button and drives 2.5 V instead

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1378,6 +1378,42 @@ int getVChipToSet(enum DACINDEX ind, int val) {
...
+
+int voltageToDac_PowerRegulators(int pwrIndex, int voltage, int *retval_dac,
+                                 char *mess) {
+    *retval_dac = -1;
+    char *powerNames[] = {PWR_NAMES};
+
+    if (ConvertToDifferentRange(POWER_RGLTR_MIN, POWER_RGLTR_MAX,
+                                LTC2620_GetMaxInput(), LTC2620_GetMinInput(),
+                                voltage, retval_dac) == FAIL) {
+        int min = pwrIndex == V_PWR_IO ? VIO_MIN_MV : POWER_RGLTR_MIN;
```
AliceMazzoleni99, line 1405: shouldnt we pass min, max to ConvertToDifferentRange?

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1378,6 +1378,42 @@ int getVChipToSet(enum DACINDEX ind, int val) {
     return max;
 }
 
+int dacToVoltage_PowerRegulators(int pwrIndex, int dac_value, int *retval_mV,
+                                 char *mess) {
+    *retval_mV = -1;
+    char *powerNames[] = {PWR_NAMES};
+    if (ConvertToDifferentRange(LTC2620_GetMaxInput(), LTC2620_GetMinInput(),
+                                POWER_RGLTR_MIN, POWER_RGLTR_MAX, dac_value,
+                                retval_mV) == FAIL) {
```
AliceMazzoleni99, line 1387: also here shouldnt we map to different values if pwrIndex == V_PWR_IO?

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1414,147 +1450,193 @@ int getADCIndexFromDACIndex(enum DACINDEX ind) {
     }
 }
 
-int isPowerValid(enum DACINDEX ind, int val) {
-    int min = (ind == D_PWR_IO) ? VIO_MIN_MV : POWER_RGLTR_MIN;
+void powerEnable(int on, int pwrIndex) {
```
AliceMazzoleni99, line 1453: on should be bool?

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1414,147 +1450,193 @@ int getADCIndexFromDACIndex(enum DACINDEX ind) {
...
              LTC2620_GetPowerDownValue()));
-        return LTC2620_GetPowerDownValue();
+        *retval = LTC2620_GetPowerDownValue();
+        return OK;
     }
 
     // vchip not set, weird error, should not happen (as vchip set to max in the
     // beginning) unless user set vchip to LTC2620_GetPowerDownValue()  and then
     // tried to get a power regulator value
     if (dacValues[D_PWR_CHIP] == -1 ||
```
AliceMazzoleni99, line 1544: what if ind != D_PWR_CHIP shouldnt we check first before doing anything else

slsDetectorServers/slsDetectorServer/include/slsDetectorFunctionList.h
```diff
@@ -373,13 +373,24 @@ int getOnChipDAC(enum ONCHIP_DACINDEX ind, int chipIndex);
 void setDAC(enum DACINDEX ind, int val, int mV, int counterEnableCheck);
 void setGeneralDAC(enum DACINDEX ind, int val, int mV);
 void setVthDac(int index, int enable);
+#elif defined(XILINX_CHIPTESTBOARDD)
+int setDAC(enum DACINDEX ind, int val, int mV);
```
AliceMazzoleni99: defined twice? e.g. #elif

slsDetectorServers/slsDetectorServer/include/slsDetectorFunctionList.h
```diff
@@ -393,15 +404,11 @@ void setVchip(int val);
 int getVChipToSet(enum DACINDEX ind, int val);
 int getDACIndexFromADCIndex(enum ADCINDEX ind);
 int getADCIndexFromDACIndex(enum DACINDEX ind);
-int isPowerValid(enum DACINDEX ind, int val);
-int getPower();
-void setPower(enum DACINDEX ind, int val);
+
 void powerOff();
 #elif XILINX_CHIPTESTBOARDD
+int getPwrIndex(enum DACINDEX ind);
```
AliceMazzoleni99: when does one use getPwrIndex and getADCIndexFromDACIndex

thattil:
> power to 0 is still a self-destruct button and drives 2.5 V instead

Yes. Apparently not setting it to any value in the config (other than 0 as 0 just disables) is like setting it to power down (-100), where it just goes to maximum value.

So, now they are updated to start up with default minimums (1041 mV for the others and 1200 for vio). Power is still disabled at startup. So a powering on the chip from the command line will set all the power regulators to the minimum values.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: I didnt get you. Is it because we pass (min V, maxV, max dac value, min dac value..)? They are inverted because thats how the conversion is in reality for LTC2620. To get the minimum value, you have to give a maximum dac value.

AliceMazzoleni99: The error message says it should be between min, max but min, max can be different than POWER_RGLTR_MIN, POWER_RGLTR_MAX

AliceMazzoleni99 (commented):
Oh woow, slsDetectorServer_funcs-cdoes loom like something. Nice that we have proper error handling now.

slsDetectorServers/slsDetectorServer/include/slsDetectorFunctionList.h
thattil: #ifdef is for m3. its hidden in this view.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: When we convert, they should be the full range it can go to. The limitation for V_pwr_io to a minimum of 1200mV  is a restriction that we are giving the users. So it should only be used ot check if its valid, but when we convert, we use the full range.

slsDetectorServers/slsDetectorServer/include/slsDetectorFunctionList.h
thattil: xilinx_ctb doesnt have adcs. so i just called it pwrindex, whereas the ctbs had them.

thattil:
replaced with PR #1389 and #1420

PR #1359, opened 2026-01-15
python setBit backwards compatible

AliceMazzoleni99:
- BitAddress only takes RegisterAddress and int as constructor arguments, convert integer first to RegisterAddress


thattil (approved):
LGTM

PR #1375, opened 2026-01-28
xilinx server: remove powerdown for power regulators

thattil:
xilinx server: not allowing power down as default dac values for the power regulators and not allowing to be set to these in the future either.. because it goes to maximum value.


AliceMazzoleni99 (approved):
LGTM.

PR #1376, opened 2026-01-28
pyctbgui decoder for Matterhorn1 16 bits (1 and 4 counters)

erikfrojdh:
- pixel map for matterhorn1 in 16 bit mode 1 or 4 counters
- decode using numpy.take
- dropdown


AliceMazzoleni99 (approved):
Looks good. Maybe its better to have a generic function 

def matterhorn1_transceiver(n_counters : int, dynamic_range : int):  
	mapping = {16: 4, 8: 8, 4: 16}  
	num_consecutive_pixels = mapping[dynamic_range]
	  for row in range(n_rows):
          for counter in range(n_counters):
              col = 0
              for offset in range(0,64,4):
                  for pkg in range(offset,256,64):
                      for pixel range(num_consecutive_pixels):
                          pixel_map[row+n_rows*counter, col] = pixel+pkg+row*n_cols*n_counters+n_cols*counter
                          col += 1

Can change afterwards.

PR #1381, opened 2026-02-04
Read and write SPI for Xilinx CTB

erikfrojdh:
Added readSpi and writeSpi to the Detector class including python bindings. The commands are intended to be wrapped in a class controlling the detector since they are very error prone when used directly. If a read of the wrong size is requested the data in the register is corrupted!

**readSpi**

There is no direct way of reading an SPI register besides shifting in data (shifts out what is in the register). Therefore a read is implemented as a write of dummy data without storing followed by a write of the data that was read out. 

**writeSpi**

No return value for write since it would return what was in the register and not what was written in. Throws if an error occurred. 

_Open questions to address later:_
- Which if any command line bindings?
- How to deal with registers that are not a multiple of a byte?


slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    if (receiveData(file_des, &n_bytes, sizeof(n_bytes), INT32) < 0){
+        return printSocketReadError();
+    }
+
+    //TODO! Make logDEBUG1 before merging
+    LOG(logINFO, ("SPI Read Requested: chip_id=%d, register_id=%d, n_bytes=%d\n",
+                    chip_id, register_id, n_bytes));
+
+
+#ifdef VIRTUAL
```
thattil, line 11151: all of this should go into the xilinx server implemetnation List.c right? Or are you waiting for me to finish the xilinx imple about dacs and power?

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    // For the virtual detector we create a fake register to read from
+    // and fill it with 0,2,4,6,... This way we can check that copying
+    // of the data works as expected
+    uint8_t *fake_register = malloc(n_bytes);
```
thattil, line 11155: One could also fake allocate all the said registers so that one can actually read and write valid values.

thattil, line 11155: check if malloc failed?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -4074,4 +4074,44 @@ void Module::simulatingActivityinDetector(const std::string &functionType,
...
+    client.Send(chip_id);
+    client.Send(register_id);
+    client.Send(n_bytes);
+
+    if (client.Receive<int>() == FAIL) {   
+        std::ostringstream os;
+        os << "Module " << moduleIndex << " (" << shm()->hostname << ")"
+           << " returned error: " << client.readErrorMessage();
+        throw DetectorError(os.str());
+    }else{
```
thattil: you dont need the 'else'.

slsDetectorSoftware/src/Module.cpp
```diff
@@ -4074,4 +4074,44 @@ void Module::simulatingActivityinDetector(const std::string &functionType,
...
+        std::vector<uint8_t> data(n_bytes);
+        client.Receive(data.data(), n_bytes);
```
thattil: I think you can even just write "client.Receive(data);"? works for std::vector<int> at least.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    for (int i = 0; i < n_bytes; i++) {
+        fake_register[i] = (uint8_t)( (i*2) % 256 ); 
+    }
+#else
+    int spifd = open("/dev/spidev2.0", O_RDWR);
```
thattil, line 11164: hardcoded path. It could be defined in one place. Ideally in xilinx server_defs.h

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+        ret = FAIL;
+        sprintf(mess, "Could not open /dev/spidev2.0\n");
+        LOG(logERROR, (mess));
+        return Server_SendResult(file_des, INT32, NULL, 0);
+    }
+#endif
+
+    // Allocate dummy data to shif in, we keep a copy of this 
+    // to double check that we access a register of the correct size
+    uint8_t *dummy_data = malloc(n_bytes);
```
thattil, line 11174: I would check if memory allocated.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    for(int i=0; i<n_bytes; i++){
+        dummy_data[i] = (uint8_t)(i % 256);
+    }
+
+    // Allocate actual data buffer this holds the data we read out 
+    // and that we need to write back to restore the register
+    uint8_t *actual_data = malloc(n_bytes);
```
thattil, line 11185: I would check if memory allocated, here and in other places.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+	send_cmd[0].tx_buf = (unsigned long) local_tx;
+	send_cmd[0].rx_buf = (unsigned long) local_rx;
+
+    // 0 - Normal operation, 1 - CSN remains zero after operation
+    // We use cs_change = 1 to not close the SPI transaction and
+    // allow for shifting the read out data back in to restore the
+    // regitster
+	send_cmd[0].cs_change = 1; 
+
+    // First byte of the message is 4 bits chip_id then 4 bits register_id
```
thattil, line 11218: all of these i wouldnt hardcode and define it in xilinx server_defs.h

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+        fake_register[i] = (uint8_t)( (i*2) % 256 ); 
+    }
+#else
+    int spifd = open("/dev/spidev2.0", O_RDWR);
+    LOG(logINFO, ("SPI Read: opened spidev2.0 with fd=%d\n", spifd));
+    if(spifd < 0){
+        ret = FAIL;
+        sprintf(mess, "Could not open /dev/spidev2.0\n");
+        LOG(logERROR, (mess));
+        return Server_SendResult(file_des, INT32, NULL, 0);
```
thattil: one should also free the allocated fake register.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    }
+
+#else
+    // For the real detector we do the transfer here
+    if(ioctl(spifd, SPI_IOC_MESSAGE(1), &send_cmd)<0){
+        ret = FAIL;
+        sprintf(mess, "SPI write failed with %d:%s\n", errno, strerror(errno));
+        LOG(logERROR, (mess));
+        close(spifd);
+        return Server_SendResult(file_des, INT32, NULL, 0);
```
thattil: free all the allocated memory before returning.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    for (int i=0; i < n_bytes; i++){
+        local_rx[i+1] = fake_register[i];
+    }
+    free(fake_register); // we are done with the fake register
+#else
+    if(ioctl(spifd, SPI_IOC_MESSAGE(1), &send_cmd)<0){
+        ret = FAIL;
+        sprintf(mess, "SPI write failed with %d:%s\n", errno, strerror(errno));
+        LOG(logERROR, (mess));
+        close(spifd);
```
thattil, line 11273: free memories before returning.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    int n_bytes = 0;
+    if (receiveData(file_des, &n_bytes, sizeof(n_bytes), INT32) < 0){
+        return printSocketReadError();
+    }
+
+    LOG(logINFO, ("SPI Write Requested: chip_id=%d, register_id=%d, n_bytes=%d\n",
+                    chip_id, register_id, n_bytes));
+    
+    uint8_t *data = malloc(n_bytes);
+    if (receiveData(file_des, data, n_bytes, OTHER) < 0)
```
thattil: return error if malloc not successful before receiveData.  free when error in receiving.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+        sprintf(mess, "Invalid register_id %d. Must be 0-15\n", register_id);
+        LOG(logERROR, (mess));
+        return Server_SendResult(file_des, INT32, NULL, 0);
+    }
+
+    int n_bytes = 0;
+    if (receiveData(file_des, &n_bytes, sizeof(n_bytes), INT32) < 0){
+        return printSocketReadError();
+    }
+
```
thattil, line 11339: if nbytes is 0, maybe return error before malloc.

slsDetectorSoftware/src/Module.cpp
```diff
@@ -4074,4 +4074,44 @@ void Module::simulatingActivityinDetector(const std::string &functionType,
...
+
+void Module::writeSpi(int chip_id, int register_id,
+                      const std::vector<uint8_t> &data){
+    auto client = DetectorSocket(shm()->hostname, shm()->controlPort);
+    client.Send(F_SPI_WRITE);
+    client.setFnum(F_SPI_WRITE);
+    client.Send(chip_id);
+    client.Send(register_id);
+    client.Send(static_cast<int>(data.size()));
+    client.Send(data.data(), data.size());
```
thattil: Maybe check if data.size() >0 and throw before sending data ? 
And you can just do 'client.Send(data);'

thattil (changes requested):
Looks good overall.

But i would move most of it to the xilinx implementation part (but maybe you were waiting on me to finish to avoid crazy conflicts?)
Malloc and free when returning and checking if malloc worked in a few places.
Sometimes malloc doesnt work for some other reason and then its better to have a proper error message than segfault.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
erikfrojdh: I would not do that since we don't know the actual number of registers. Then we need to define where register 1 stops and where register 2 starts etc. If we decide to add defines for spi then we can also add support in the virtual server.

slsDetectorSoftware/src/Module.cpp
erikfrojdh: Removed

slsDetectorSoftware/src/Module.cpp
erikfrojdh: True, since we allocate the data for the vector...

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,260 @@ int set_pattern_wait_interval(int file_des) {
...
+    local_tx[0] = ((chip_id & 0xF) << 4) | (register_id & 0xF);
+	for (int i=0; i < n_bytes; i++)
+		local_tx[i+1] = actual_data[i];
+
+#ifdef VIRTUAL
+    // Copy the data from the fake register to the local_rx buffer
+    for (int i=0; i < n_bytes; i++){
+        local_rx[i+1] = fake_register[i];
+    }
+    free(fake_register); // we are done with the fake register
```
AliceMazzoleni99, line 11269: do we even need to write to local_rx?

AliceMazzoleni99 (commented):
Looks good to me. Nice to support behavior for virtual servers would have forgotten about it.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
erikfrojdh: Not here since we only try to open SPI on the real detector

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
erikfrojdh: As the code looks now, not really but if we later want to add a comparison before returning it's good. Put it there to be equivalent to the real detector since ioctl would write to local_rx

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
erikfrojdh: Done!

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
erikfrojdh: Added a print and EXIT_FAILURE, for small allocations it doesn't look like we get the error message back, rather the server crashes anyway. With doing the cleanup progressively of all allocations we would get quite some extra lines. (or one need to move all allocations to the top)

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
erikfrojdh: In step two when moving the implementation

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
thattil: Makes sense. 👍

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,299 @@ int set_pattern_wait_interval(int file_des) {
...
+        return Server_SendResult(file_des, INT32, NULL, 0);
+    }
+
+    int n_bytes = 0;
+    if (receiveData(file_des, &n_bytes, sizeof(n_bytes), INT32) < 0){
+        return printSocketReadError();
+    }
+    if(n_bytes < 1 ){
+        sprintf(mess, "Invalid n_bytes %d. Must ask for a write of at least 1 byte\n", n_bytes);
+        return sendError(file_des);
```
thattil, line 11337: minor: log(logERROR, (mess) missing.

erikfrojdh: is done by sendError

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,299 @@ int set_pattern_wait_interval(int file_des) {
...
+	for (int i=0; i < n_bytes; i++)
+		local_tx[i+1] = data[i];
+
+#ifdef VIRTUAL
+    // For the virtual detector we have nothing to do
+#else
+    int spifd = open("/dev/spidev2.0", O_RDWR);
+    LOG(logINFO, ("SPI Read: opened spidev2.0 with fd=%d\n", spifd));
+    if(spifd < 0){
+        sprintf(mess, "Could not open /dev/spidev2.0\n");
```
thattil, line 11386: free before sending error?

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11095,4 +11103,299 @@ int set_pattern_wait_interval(int file_des) {
...
+        return sendError(file_des);
+    }
+    if(ioctl(spifd, SPI_IOC_MESSAGE(1), &send_cmd)<0){
+        close(spifd);
```
thattil, line 11390: same here. free?

thattil (commented):
Just 1 fix and 1 minor fix:
- server_funcs: free before returning error in spi_write
- Modeule: client.send(data) for writespi (minor) just looks easier

thattil (approved):

PR #1385, opened 2026-02-06
skip if not plotted

### LINKED ISSUE #1378: CtbGui does not display DBIT graph when only selected DBITs are enabled
When DBITs are acquired in the CtbGui they are only updated in the display if all bits are selected for the plot. When only a single DBit is shown in the plot it displays the data from the last acquisition and does not update.

AliceMazzoleni99:
- cant set np.nan for an integera numpy array (setting to zero) 
- instead of checking data - skip if plot not enabled 
- unchecks plot box if bit disabled
- always write and store all the acquisition data (even if plot not enabled)

### DIFF
--- a/pyctbgui/pyctbgui/services/Signals.py
+++ b/pyctbgui/pyctbgui/services/Signals.py
@@ -112,7 +112,7 @@ class SignalsTab(QtWidgets.QWidget):
                     bit_index += (8 - (bit_index % 8))
                 if not isPlottedArray[i]:
                     bit_index += nbitsPerDBit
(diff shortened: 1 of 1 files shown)


thattil (changes requested):
- For digital plots, the first time you turn on the gui, take an acquisition, enable plots, it wont show, it shows only the legend. but once you have clicked on the enable plots at least once. next time you acquire, it will show. Is this to do with self.plots.addDigitalPlots is only inside setBitPlot and not at the startup maybe?

thattil (approved):

PR #1389, opened 2026-02-10
Dev/xilinx set dac rewrite

thattil:
- set_dac in  in slsDetectorFuncs.c has become unmaintainable and unreadable
- separating the generic set_dac in slsDetectorFuncs.c that does highvoltage, setpower among other things including setdac (normal dacs). ( cannot call different function enums in Module.cpp yet as it will break compatibility)
- passing the ret and mess into the individual detector implementations so that they can pass the direct issues immediately.
- setting power dacs in setPower should call setDAC directly and handle conversions inside.
- writeDACspi should only handle dac values and writing to file or bit banging, but is the low level function.


slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -587,15 +589,24 @@ void setupDetector() {
...
+    // dacs only
     LOG(logINFOBLUE, ("Powering down all dacs\n"));
-    for (int idac = 0; idac < NDAC; ++idac) {
-        setDAC(idac, LTC2620_GetPowerDownValue(),
-               0); // has to be before setvchip
+    for (int idac = 0; idac < NDAC_ONLY; ++idac) {
+        initError =
+            setDAC(idac, LTC2620_GetPowerDownValue(), 0, initErrorMessage);
+        if (initError == FAIL)
+            return;
```
AliceMazzoleni99, line 599: should the function return int? and we return EXIT_FAILURE?

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,338 +1236,543 @@ int64_t getMeasurementTime() {
 enum detectorSettings getSettings() { return UNDEFINED; }
 
 /* parameters - dac, adc, hv */
+int setADCVpp(int val, int mV, char *mess) {
```
AliceMazzoleni99, line 1243: mV should be bool. Maybe also add a docstring for new  added functions.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: We aren't exiting the server with a failure here, but the error message is saved so that the client can access it the next time it tries to connect to it.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,338 +1236,543 @@ int64_t getMeasurementTime() {
...
 
-int getVChipToSet(enum DACINDEX ind, int val) {
+int getVChipToSet(enum DACINDEX ind, int val, int *vchip, char *mess) {
     LOG(logDEBUG1, ("Calculating vchip to set\n"));
-    // validate index & get adc index
-    int adcIndex = getADCIndexFromDACIndex(ind);
-    if (adcIndex == -1) {
-        return -1;
-    }
+    enum PWRINDEX pwrIndex = 0;
```
AliceMazzoleni99: minot: pwrindex = PWR_I0

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,338 +1236,543 @@ int64_t getMeasurementTime() {
...
+    char *powerNames[] = {PWR_NAMES};
 
-    // get maximum value of the adc values (minimum is 0)
+    // validate index
+    if (getPowerIndex(ind, &pwrIndex, mess) == FAIL)
```
AliceMazzoleni99: maybe better name DACIndex2PowerIndex

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,338 +1236,543 @@ int64_t getMeasurementTime() {
...
+    return OK;
+}
 
-            // set and power on/ update dac
-            LOG(logINFO, ("Setting P%d (DAC %d): %d dac (%d mV)\n", adcIndex,
-                          ind, dacval, val));
-            setDAC(ind, dacval, 0);
+int initPower(enum DACINDEX ind, char *mess) {
+    if (ind == D_PWR_CHIP)
+        return OK;
```
AliceMazzoleni99: for me its not really clear whats the difference between initPOwer and setPower

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1720,24 +1938,33 @@ int getSlowADCTemperature() {
...
-            bus_w(addr, bus_r(addr) | POWER_HV_INTERNAL_SLCT_MSK);
+    // switch on internal high voltage, if set
+    if (val > 0)
+        bus_w(addr, bus_r(addr) | POWER_HV_INTERNAL_SLCT_MSK);
 
-        highvoltage = val;
-    }
-    return highvoltage;
+    highvoltage = val;
+    return OK;
```
AliceMazzoleni99, line 1881: when is highvoltage used?

slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1563,61 +1614,85 @@ int getADC(enum ADCINDEX ind) {
...
+        LOG(logERROR, ("Could not read high voltage\n"));
+        sharedMemory_unlockLocalLink();
+        strcpy(
+            mess,
+            "Getting high voltage failed. Serial/i2c communication failed.\n");
+        LOG(logERROR, (mess));
+        return FAIL;
+    }
+    // need to read the file twice to get the proper value
+    if (!Feb_Control_GetHighVoltage(&eiger_highvoltage)) {
```
AliceMazzoleni99, line 1659: why is eiger_highvoltage a global variable? same for eiger_theo_highvoltage

slsDetectorServers/slsDetectorServer/src/DAC6571.c
```diff
@@ -94,9 +102,10 @@ int DAC6571_Get(int *retval) {
     // open file
     FILE *fd = fopen(DAC6571_DriverFileName, "w");
     if (fd == NULL) {
-        LOG(logERROR,
-            ("Could not open file %s for writing to set high voltage\n",
-             DAC6571_DriverFileName));
+        sprintf(mess,
```
AliceMazzoleni99, line 105: why do we have to write for virtual detectors for get?

slsDetectorServers/slsDetectorServer/src/LTC2620_Driver.c
```diff
@@ -69,74 +68,30 @@ int LTC2620_D_SetDACValue(int dacnum, int val, int mV, char *dacname,
         return FAIL;
     }
 
-    // validate set
-    if (val < 0 && val != LTC2620_D_PWR_DOWN_VAL)
+    // validate value
+    if ((val < 0 && val != LTC2620_D_PWR_DOWN_VAL) ||
```
AliceMazzoleni99: mmh dont we check in validateDAC is it universal for all detectors?

slsDetectorServers/slsDetectorServer/src/LTC2620_Driver.c
```diff
@@ -150,16 +105,12 @@ int LTC2620_D_SetDACValue(int dacnum, int val, int mV, char *dacname,
...
 #ifdef XILINX_CHIPTESTBOARDD
-    // not changing *dacval from -100 (cant write -100 to file: invalid arg)
-    int writeValue = *dacval;
-    if (writeValue == LTC2620_D_PWR_DOWN_VAL)
-        writeValue = 1;
-    fprintf(fd, "%d\n", writeValue);
-#else
-    fprintf(fd, "%d\n", *dacval);
+    if (val == LTC2620_D_PWR_DOWN_VAL)
+        fprintf(fd, "1\n"); //-100?:Invalid Argument
```
AliceMazzoleni99: mmh but we have written to LTC2620_D_PowerDownDriverFileName) anyway for xilinx we never write to  LTC2620_D_DriverFileName?

slsDetectorServers/slsDetectorServer/src/MAX1932.c
```diff
@@ -43,33 +43,33 @@ void MAX1932_Disable() {
...
     // limit values (normally < 60 => 0 (off))
-    if (*val < MAX1932_MinVoltage) {
-        dacvalue = MAX1932_POWER_OFF_DAC_VAL;
-        *val = 0;
-    }
-    // limit values (normally > 200 => 0x1 (max))
-    else if (*val > MAX1932_MaxVoltage) {
-        dacvalue = MAX1932_MAX_DAC_VAL;
-        *val = MAX1932_MaxVoltage;
+    if ((val != 0 && val < MAX1932_MinVoltage) || val > MAX1932_MaxVoltage) {
```
AliceMazzoleni99, line 50: again shouldnt we check this before setting high voltage?

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
AliceMazzoleni99, line 1: maybe its even better to have seperate functions validateAndSetDac for each detector instead of all the if defs

AliceMazzoleni99: But with the refactoring its already so much cleaner - thats really nice

AliceMazzoleni99 (commented):
Nice work. Its already so much cleaner. I think it would really improve readablity even more if conversion between mv and dac unit would be seperated instead of doing it on the go in the functions. So the validation also becomes clearer.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: Wasn't aware we could use bool before in c. So this is quite awesome. I have made this change now.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: it was just to set only dacs without touching power enable. But I have rewritten and removed thsi initPower function and instead I call setDAC directly. So i hope its more clearer this time.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: only to get the value that was set before.

slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
thattil: good catch. Removed eiger_highvoltage as a global variable. The eiger_theo_highvotlage has to be global though because it is set in SetHighVotlage and then used in GetHighVoltage.

slsDetectorServers/slsDetectorServer/src/LTC2620_Driver.c
thattil: not the same for the power dacs. but this code has been changed now.

slsDetectorServers/slsDetectorServer/src/LTC2620_Driver.c
thattil: we do. Its used in the fnameFormat. to get the fname.

slsDetectorServers/slsDetectorServer/src/MAX1932.c
thattil: The min and max of MAX1932 are defined here. So it makes sense to me to check this here and complain if it doesnt fit.

thattil:
> Nice work. Its already so much cleaner. I think it would really improve readablity even more if conversion between mv and dac unit would be seperated instead of doing it on the go in the functions. So the validation also becomes clearer.

Done

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -587,15 +589,28 @@ void setupDetector() {
...
     }
 
     // power regulators
+    LOG(logINFOBLUE,
+        ("Setting power dacs to min dac value (power disabled)\n"));
+    for (int idac = NDAC_ONLY; idac < NDAC; ++idac) {
+        if (idac == D_PWR_CHIP)
+            continue;
+        int min = (idac == D_PWR_IO) ? VIO_MIN_MV : POWER_RGLTR_MIN;
+        initError = setDAC(idac, min, true, initErrorMessage);
```
AliceMazzoleni99, line 609: maybe this should call setPower?

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,345 +1240,500 @@ int64_t getMeasurementTime() {
...
+    return OK;
 }
 
-int getMaxDacSteps() { return LTC2620_GetMaxNumSteps(); }
+int validateDACValue(enum DACINDEX ind, int dacval, char *mess) {
+    // validate min value
+    if (dacval < 0 && dacval != LTC2620_GetPowerDownValue()) {
+        sprintf(
+            mess,
+            "Could not set DAC %d. Input value %d must be positive or -100.\n",
```
AliceMazzoleni99: maybe better to use LTC2620_GetPowerDownValue() instead of hardcoding -100

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,345 +1240,500 @@ int64_t getMeasurementTime() {
...
 
-int getVChipToSet(enum DACINDEX ind, int val) {
-    LOG(logDEBUG1, ("Calculating vchip to set\n"));
-    // validate index & get adc index
-    int adcIndex = getADCIndexFromDACIndex(ind);
-    if (adcIndex == -1) {
-        return -1;
-    }
+int writeDACSpi(enum DACINDEX ind, int dacval, char *mess) {
+    if (validateDACValue(ind, dacval, mess) == FAIL)
```
AliceMazzoleni99: maybe we should not validate in here at all - but only when setting the DAC.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,345 +1240,500 @@ int64_t getMeasurementTime() {
...
+    return OK;
+}
+
+int setVchip(int val, char *mess) {
+    LOG(logINFOBLUE, ("Setting Vchip to %d mV\n", val));
+    if (validateVchip(val, mess) == FAIL)
+        return FAIL;
+
+    if (setDAC(D_PWR_CHIP, val, true, mess) == FAIL)
+        return FAIL;
```
AliceMazzoleni99, line 1475: mmh maybe again less error prone if directly written to SPI using writeDACSPI. Conversion to dac can happen in here.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,345 +1240,500 @@ int64_t getMeasurementTime() {
...
+        } else {
+            if (getPower(ind, &val, mess) == FAIL)
+                return FAIL;
+        }
+        if (val > max)
+            max = val;
+    }
     // increment to get vchip value
-    max += VCHIP_POWER_INCRMNT;
+    *retval_vchip = max + VCHIP_POWER_INCRMNT;
```
AliceMazzoleni99, line 1500: could be tricky to set it before returning FAIL

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: No because it should not enable power. It should only set the dacs values to minimum. If I call setPower, it will set dac and enable the power since its greater than 0.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1223,345 +1240,500 @@ int64_t getMeasurementTime() {
...
+int getPower(enum DACINDEX ind, int *retval, char *mess) {
+    enum PWRINDEX pwrIndex = PWR_IO;
+    if (getPowerIndexFromDACIndex(ind, &pwrIndex, mess) == FAIL)
+        return FAIL;
 
-        // power down dac
-        LOG(logDEBUG1, ("Powering off P%d (DAC %d)\n", adcIndex, ind));
-        setDAC(ind, LTC2620_GetPowerDownValue(), 0);
+    // if powered off, return 0
+    if (getPowerRail(pwrIndex, retval, mess) == FAIL)
```
AliceMazzoleni99, line 1656: why do we need that e.g. call getPowerRail?

slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1422,101 +1426,198 @@ int setThresholdEnergy(int ev) {
 }
 
 /* parameters - dac, adc, hv */
+int validateDACIndex(enum DACINDEX ind, char *mess) {
+    // including E_VTHRESHOLD
+    if (ind < 0 || ind >= NDAC + 1) {
```
AliceMazzoleni99, line 1431: why indexing is from 0 - 23?

slsDetectorServers/mythen3DetectorServer/slsDetectorFunctionList.c
```diff
@@ -1543,89 +1539,262 @@ void setThresholdEnergy(int counterIndex, int eV) {
...
     const int specialDacs[NSPECIALDACS] = SPECIALDACINDEX;
     for (int i = 0; i < NSPECIALDACS; ++i) {
         if ((int)ind == specialDacs[i]) {
             validateSettings();
         }
     }
+    return OK;
+}
+
+int getCounterIndex(enum DACINDEX ind) {
```
AliceMazzoleni99: minor: maybe also rename to fromDacIndexToCounterIndex to stay consistent

slsDetectorServers/mythen3DetectorServer/slsDetectorFunctionList.c
```diff
@@ -1543,89 +1539,262 @@ void setThresholdEnergy(int counterIndex, int eV) {
...
+    switch (ind) {
+    case M_VTH1:
+        return 0;
+    case M_VTH2:
+        return 1;
+    case M_VTH3:
+        return 2;
+    default:
+        return -1;
```
AliceMazzoleni99: minor: return FAIL as default is better

AliceMazzoleni99: maybe even passing char string for error message handling error in here.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: Moving the dacVAlue checks into LTC, where it exists anyway as a final check.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: I dont think so. because it has to convert from mV to dac units and this we do inside setDAC now.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: I didnt understand sorry. It is not setting anything in this function. It is only computing vchip to set and validating in this function.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: if the power rail is disabled, then we dont need to read the dac value because it would be 0.

slsDetectorServers/eigerDetectorServer/slsDetectorFunctionList.c
thattil: In Eiger, NDAC= 16, not 24. So here, the index allowed is from 0-16 because of vthreshold dac (just like m3). That sets 5 dacs together for Eiger.

slsDetectorServers/mythen3DetectorServer/slsDetectorFunctionList.c
thattil: changed to getCounterIndexFromDacIndex

slsDetectorServers/mythen3DetectorServer/slsDetectorFunctionList.c
thattil: I can't do that because FAIL is 1, which is a valid result at this point. I can do the error message.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
AliceMazzoleni99: Its only a minor. I would store in a temporary. max + VCHIP_POWER_INCRMT and set retval_chip after the last possible return of FAIL

AliceMazzoleni99 (approved):

PR #1391, opened 2026-02-10
dev/handes trailing / in filepath 

### LINKED ISSUE #1386: FilePath with trailing / is not handled well
##### *Distribution: 

debian 12

##### *Detector type: 

Jungfrau, but should apply to all others as well

10.0.0

##### Priority:

Low

##### *Describe the bug

On many places in the code it is assumed, that the filePath has no trailing "/", but such a string is taken without complaints. This results in double slashes for example in the fname attribute of the ZMQ stream. As path names are often given with a trailing /, this is unfortunate.

##### Expected behavior

Double / should never appear in a path string. So a respective user input should be handled to avoid this.

##### To Reproduce

Call the API function setFilePath with a parameter like "/var/temp/test/" - anything with a trailing "/". Enable streaming, start an acquisition and have a look into the fname attribute of the ZMQ stream.

##### Additional context

I didn't dive deep enough into the code to be sure it's the right place to do this, but the following patch would resolve the issue:

slsDetectorPackage-10.0.0/slsReceiverSoftware/src/Implementation.cpp
489a490,491
>     while(filePath.back() == '/')
>         filePath.pop_back();

AliceMazzoleni99:
- filePath member in Receiver Implementation is of type std::filesystem::path

### DIFF
--- a/slsReceiverSoftware/src/Implementation.cpp
+++ b/slsReceiverSoftware/src/Implementation.cpp
@@ -16,7 +16,6 @@
 #include <cstdlib> //system
 #include <cstring>
 #include <cstring> //strcpy
-#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <sys/stat.h> // stat
@@ -656,7 +655,7 @@ void Implementation::startReceiver() {
                 generalData->dynamicRange,
                 numPorts,
                 static_cast<size_t>(generalData->imageSize),
-                filePath,
+                filePath.string(),
                 fileName,
                 fileIndex,
                 quadEnable,
@@ -856,9 +855,7 @@ void Implementation::ResetParametersforNewAcquisition() {
         it->ResetParametersforNewAcquisition();
 
     if (dataStreamEnable) {
-        std::ostringstream os;
-        os << filePath << '/' << fileName;
-        std::string fnametostream = os.str();
+        std::string fnametostream = (filePath / fileName).string();
         for (const auto &it : dataStreamer)
             it->ResetParametersforNewAcquisition(fnametostream);
     }
@@ -880,6 +877,10 @@ void Implementation::SetupWriter() {
 
     try {
         // check if folder exists and throw if it cant create
+        if (filePath.empty()) {
+            throw RuntimeError("File path cannot be empty. Please set file "
+                               "path before starting acquisition.");
+        }
         if (!std::filesystem::exists(filePath)) {
             try {
                 std::filesystem::create_directories(filePath);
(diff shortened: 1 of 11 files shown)


slsReceiverSoftware/src/Implementation.cpp
```diff
@@ -16,6 +16,7 @@
 #include <cstdlib> //system
 #include <cstring>
 #include <cstring> //strcpy
+#include <filesystem>
```
thattil: Do you need this include here and in other places? This already includes "Implementation.h", which already includes the filesystem header.

slsReceiverSoftware/src/Implementation.cpp
```diff
@@ -874,21 +879,24 @@ void Implementation::CreateUDPSockets() {
 void Implementation::SetupWriter() {
 
     try {
```
thattil, line 878: I think it is safer on all platforms if one still tries  to check if the path is empty.  If the user never sets it, it will be empty when it calls SetupWriter.

slsReceiverSoftware/src/Implementation.h
```diff
@@ -316,7 +317,7 @@ class Implementation : private virtual slsDetectorDefs {
 
     // file parameters
     fileFormat fileFormatType{BINARY};
-    std::string filePath{};
+    std::filesystem::path filePath{};
```
thattil, line 320: filePath is used as a string to  many functions in Implementation.cpp. I doubt it is being converted to string everywhere. For eg. CreateMasterFile,  startACquiistionCallback, ResetParameters..

thattil (changes requested):
Looks better using filesystem instad of string for file path! I think there are a few places its missed out in Implementation.cpp though.

slsReceiverSoftware/src/Implementation.cpp
AliceMazzoleni99: removed all the unneccessary ones

slsReceiverSoftware/src/Implementation.cpp
AliceMazzoleni99: Done.

slsReceiverSoftware/src/Implementation.h
AliceMazzoleni99: Good point! Checked everywhere that filePath is used correctly.

slsReceiverSoftware/src/MasterFileUtility.cpp
```diff
@@ -336,7 +337,15 @@ std::string CreateVirtualHDF5File(
                     os << filePath << "/" << fileNamePrefix << "_d"
```
thattil: this "os" variable is unused. I believe this is what you are replacing..

thattil (approved):

PR #1405, opened 2026-02-26
xilinx: slow adcs

### LINKED ISSUE #1402: slow ADC on the XCTB board
As the old CTB board, also the XCTB board has some slow ADCs.
They can be readout with "slowadc", but they return an incorrect value. However the value is not random, but proportional to the input voltage, so in python I multiply by an approximate empirical constant to get the value in mV or uV. It would be great to put in the function the correct conversion coefficient for that ADC and the corresponding reference voltage.

thattil:
Slow adcs now convert dac units to uV (also requires change in device tree to set internal ref to 2.5V and not 4.096V


AliceMazzoleni99 (commented):
Looks good. Maybe you have to update the API version of xilinx.

thattil:
> Looks good. Maybe you have to update the API version of xilinx.

Done. It wasnt working until now because the script wasnt updated to reflect etc/updateVersionAPI.py. I did it in a separate PR merged to developer.

PR #1406, opened 2026-02-27
Dev/update workflows

AliceMazzoleni99:


.gitea/workflows/docker-rh8-build-test.yml
```diff
@@ -26,4 +26,17 @@ jobs:
```
thattil, line 26: I would include "rh8 docker" or something like that in the name of the workflow.  Similar to how you have changed the name of the file. Minor.

.gitea/workflows/rh8-local-build-deploy.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             conda activate det
             mkdir build && cd build
             conda activate det
-            cmake .. -DSLS_USE_PYTHON=ON
+            cmake .. -DSLS_USE_PYTHON=ON -DSLS_USE_SIMULATOR=ON -DSLS_USE_GUI=ON
```
thattil, line 24: 👍

thattil (approved):
Just minor. Name of workflow  is the only thing left. Otherwise looks good to me.

PR #1413, opened 2026-03-04
Adding offset to RegisterAddress

### LINKED ISSUE #1412: Add offset to RegisterAddress
```cpp
RegisterAddress a(5);
int b = 7;

auto c = a + b; // c is a RegisterAddress
```

erikfrojdh:
1. Added operator + and += to RegisterAddress for uint32_t
2. Added tests
3. Added python bindings and tests
4. Fixed failing tests and removed [.bit_utils] which was preventing them from running

Additional fix for failing shm test when SLSDETNAME was set.

### DIFF
--- a/python/tests/test_bits.py
+++ b/python/tests/test_bits.py
@@ -1,4 +1,5 @@
 from slsdet.bits import clearbit, clearbit_arr, setbit, setbit_arr
+from slsdet import RegisterAddress
 import numpy as np
 
 
@@ -48,3 +49,12 @@ def test_clearbit_arr():
     arr = np.array((5, 5, 5), dtype=np.int8)
     clearbit_arr(0, arr)
     assert all(arr == (4, 4, 4))
+
+def test_RegisterAddress_addition():
+    r = RegisterAddress(0x10)
+    r2 = r + 0x5
+    assert r2.value() == 0x15
+    r3 = 0x5 + r
+    assert r3.value() == 0x15
+    r += 0x5
+    assert r.value() == 0x15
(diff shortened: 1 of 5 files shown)


slsSupportLib/include/sls/bit_utils.h
```diff
@@ -102,4 +111,9 @@ std::ostream &operator<<(std::ostream &os, const RegisterAddress &r);
 std::ostream &operator<<(std::ostream &os, const BitAddress &r);
 std::ostream &operator<<(std::ostream &os, const RegisterValue &r);
 
+constexpr RegisterAddress operator+(uint32_t offset,
```
thattil, line 114: ah that was good to know :)

slsDetectorSoftware/tests/test-SharedMemory.cpp
```diff
@@ -42,9 +42,11 @@ constexpr int shm_id = 10;
 // macOS does not expose shm in the filesystem
 #ifndef __APPLE__
 
+const char *env_p = std::getenv(SHM_ENV_NAME);
```
thattil, line 45: good catch!

python/tests/test_bits.py
```diff
@@ -47,4 +48,13 @@ def test_setbit_arr():
 def test_clearbit_arr():
     arr = np.array((5, 5, 5), dtype=np.int8)
     clearbit_arr(0, arr)
```
thattil, line 50: ah there are tests missing for the others. I think I had added tests with "make_RegisterAddress" or similar in test_utils with the original PR, but it got lost when we removed the make. We should probably add other bits.cpp tests in here like you did for RegisterAddress. Do you want to do it or shall I do it in another PR?

thattil (approved):

PR #1414, opened 2026-03-05
added support for int() for RegisterAddress and RegisterValue

erikfrojdh:
Added support for:

```python
from slsdet import RegisterAddress, RegisterValue
a = RegisterAddress(0x5)
i = int(a)

v = RegisterValue(32)
i = int(v)
```

Uses the .value() conversion


AliceMazzoleni99 (approved):
Nice Addition

PR #1415, opened 2026-03-05
rm unnecessary hidden test tags

thattil:
other than .detectorintegration and .disable_check_data_file


slsSupportLib/tests/test-bit_utils.cpp
```diff
@@ -121,8 +121,7 @@ TEST_CASE("Convert BitAddress using classes", "[support]") {
     }
 }
 
-TEST_CASE("Output operator gives same result as string",
-          "[support]") {
+TEST_CASE("Output operator gives same result as string", "[support]") {
     {
```
AliceMazzoleni99, line 126: should it have label reg?

AliceMazzoleni99 (approved):

PR #1420, opened 2026-03-11
Dev/ctb separate dac and power

thattil:
Main features:
- powerIndex enum => getPowerIndex, set/getPowerName, type changed from dacIndex
- removed v_a, v_b, .. v_io, v_chip (new file->removed commands like deprecated commands)
- separated dac and enable part of power (power and powerdac) commands (api change => setPowerDAc, setPowerEnabled) 
- possible to enable a list of powers 
- vchip in getPowerdac only
- vlimit has its own function=> get/setVoltageLimit (not reusing setDAC), default vlimit = 0, setting any dac/power >0, validated with vlimit
- tests added for python commands
- removed powerchip from ctb and xilinx ctb (suing boolean here in the servers)
- command generation for all the powe indices
- command generation for remvoed commands (does not execute anything else, just suggest)

Side Fix:
- eiger detector mac address on localhost. fix
- Bechir's scripts needs c++17 updated

Details in python:
- ctb removed from comments in powerchip command
- removed v_a..v_io
- moved slowadc, slowadclist, adclist, signallist, powerlist into Ctb
- ctb only uses NamedDacs (can change dac names, only for single module), 
- other detectors use DetectorDacs (cannot change dac names, for multi modules)
- 'DetectorPowers' ->'NamedPowers' to be consistent that 
   - only ctb uses it
   - one can name it and 
   - its only for single mods
-  frozen: removed it as a class variable and made it an instance variable
- NamedDacs and NamedPowers: 
   - removed dacnames and powernames as fixed attributes and made them variables picked up from powernames everytime (to reflect latest changes)
   - __next__ uses getattr instead of __getattr__ because the power names and dac names are not attributes anymore
- Dacs:
   - changed max from 4000 to 4096 and default from 1000 to -100
- Powers:
   - removed power values as powers does the same thing
   - does not inherit DetectorProperty anymore because powers are not multi module anymore in detector api
   - removed all the multi module array setting features
   - can set 'dac' and 'enable' separately in powers.VA for example. 
- tests
   - added tests for v_limit, v_abcd, powers, adclist, signallist, sloweadc, dac


python/slsdet/ctb.py
```diff
@@ -24,4 +27,66 @@ def dacs(self):
 
     @property
     def powers(self):
-        return self._powers
\ No newline at end of file
+        return self._powers
+    
+    @property
+    def powerlist(self):
```
AliceMazzoleni99, line 33: maybe rename from powerlist to powernames

python/slsdet/dacs.py
```diff
@@ -41,51 +41,47 @@ class NamedDacs:
...
+        if hasattr(self._detector, 'daclist'):
+            return [n.replace(" ", "") for n in self._detector.daclist]
         else:
-            self._dacnames = [n.replace(" ", "") for n in detector.getDacNames()]
-
-        # Populate the dacs
-        for i,name in enumerate(self._dacnames):
-            #name, enum, low, high, default, detector
-            setattr(self, name, Dac(name, dacIndex(i), 0, 4000, 1000, detector))
+            return [f"DAC{i}" for i in range(18)]
```
AliceMazzoleni99: but doesnt every detector have a daclist - why hardcode to size of 18?

python/slsdet/ctb.py
thattil: The reason its powerlist is because it goes with all the other "lists" we have.. we have settingslist, daclist, adclist, drlist,timinglist, readoutspeedlist etc.. i think we should keep it for consistency.  Or we can change all the "setPowerNames, setAdcNames, setDACnames (all plural) to setPowerList etc.. but then I would do this in another PR.

python/slsdet/dacs.py
thattil: The old code was using hardcoded values if detector size is 0. I think Ill just replace the whol _dacnames with      @property
    def _dacnames(self):
        if self._detector.size() == 0:
            raise RuntimeError("No modules added")
        return [n.replace(" ", "") for n in self._detector.daclist]

I have also done something ismilar for _powernames in powers.py

python/slsdet/powers.py
```diff
@@ -1,195 +1,119 @@
...
-        self._frozen = True
+    @property
+    def _powernames(self):
+        if self._detector.size() == 0:
+            raise RuntimeError("No modules added")
+        # always get the latest list
+        if hasattr(self._detector, 'powerlist'):
+            return [n.replace(" ", "") for n in self._detector.powerlist]
+        else:
+            return ["VA", "VB", "VC", "VD", "VIO"]
```
AliceMazzoleni99: what is with I_POWER_A etc.

python/src/detector.cpp
```diff
@@ -1766,18 +1776,19 @@ void init_det(py::module &m) {
                            Detector::getPowerNames);
     CppDetectorApi.def(
         "getPowerIndex",
-        (defs::dacIndex(Detector::*)(const std::string &) const) &
+        (defs::powerIndex(Detector::*)(const std::string &) const) &
```
AliceMazzoleni99, line 1779: minor: should this be renamed to power enum?

python/tests/test_det_api.py
```diff
@@ -394,3 +394,320 @@ def test_patternstart(session_simulator, request):
...
+    """Test powers and powerlist."""
+    det_type, num_interfaces, num_mods, d = session_simulator
+    assert d is not None
+
+    from slsdet import Ctb
+    c = Ctb()
+    
+    if det_type in ['ctb', 'xilinx_ctb']:
+
+        c.powerlist
```
AliceMazzoleni99, line 490: what should this do

slsDetectorSoftware/include/sls/Detector.h
```diff
@@ -1629,20 +1630,49 @@ class Detector {
     Result<int> getSYNCClock(Positions pos = {}) const;
 
     /** gets list of power enums */
-    std::vector<defs::dacIndex> getPowerList() const;
+    std::vector<defs::powerIndex> getPowerList() const;
 
-    /** gets list of slow adc enums */
-    std::vector<defs::dacIndex> getSlowADCList() const;
+    /** [CTB] Options: V_POWE_A, V_POWER_B, V_POWER_C,
```
AliceMazzoleni99: typo: V_POWER_A

slsDetectorSoftware/src/Caller.h
```diff
@@ -902,6 +899,14 @@ class Caller {
         {"frameindex", "rx_frameindex"},
 
     };
+
+    StringMap removed_functions{
+
+        {"v_a", "dac and power"},  {"v_b", "dac and power"},
+        {"v_c", "dac and power"},  {"v_d", "dac and power"},
+        {"v_io", "dac and power"}, {"v_chip", "dac"},
```
AliceMazzoleni99: maybe add dac v_a and power v_a?

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -2398,7 +2400,13 @@ defs::dacIndex Detector::getDacIndex(const std::string &name) const {
             throw RuntimeError("Dac name not found");
         return static_cast<defs::dacIndex>(it - names.begin());
     }
-    return StringTo<defs::dacIndex>(name);
+    auto retval = StringTo<defs::dacIndex>(name);
+    auto list = getDacList();
+    if (std::find(list.begin(), list.end(), retval) == list.end()) {
+        throw RuntimeError("Dac name not found in dac list. Use 'daclist' to "
+                           "get the list of dac names for this detector.");
```
AliceMazzoleni99: maybe rename to get the list of dac indices?

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -2408,6 +2416,11 @@ void Detector::setDacName(const defs::dacIndex i, const std::string &name) {
 std::string Detector::getDacName(const defs::dacIndex i) const {
     if (pimpl->isChipTestBoard())
         return pimpl->getCtbDacName(i);
+    auto list = getDacList();
+    if (std::find(list.begin(), list.end(), i) == list.end()) {
+        throw RuntimeError("Dac index not found in dac list. Use 'daclist' to "
+                           "get the list of dac indices for this detector.");
+    }
```
AliceMazzoleni99, line 2426: use daclist? Not sure if intended.  getDacList always returns enum indices , getDacNames names if ctb enum indices for all others. Caller daclist always returns getDacNames? so daclist never returns enum indices for ctb?

python/slsdet/powers.py
thattil: thats only for measured current, we reuse for measured power. its for getPowerADC. powers is only to set and get the power (enable and dacs). Its not for measured values. For that, the commands are vm_a, vm_b... im_a, im_b..etc

python/src/detector.cpp
thattil: Yes. Good catch! I guess I should run the pybind script again.

AliceMazzoleni99 (commented):
Looks good to me. Also the python part looks good. Added some minor comments.

python/src/detector.cpp
thattil: Thats strange. It is powerIndex and not dacIndex. Maybe you should refresh?

python/tests/test_det_api.py
thattil: It should just not throw. This command should exist and execute without errors. It should get list of powers.

slsDetectorSoftware/include/sls/Detector.h
thattil: Done

python/slsdet/dacs.py
AliceMazzoleni99: but couldnt you just always return daclist?

python/slsdet/ctb.py
AliceMazzoleni99: Ah oke lets keep it for now.

slsDetectorSoftware/src/Caller.h
thattil: Done

python/src/detector.cpp
AliceMazzoleni99: All fine I was confused at first that index is the enum value

slsDetectorSoftware/src/Detector.cpp
thattil: I have changed the error messages for both getDacName and getDacIndex. As discussed, we might need a separate PR to move the adclist, signallist etc into adcnames, signalnames etc and in command line and python, solve for daclist/dacnames

AliceMazzoleni99 (approved):

PR #1423, opened 2026-03-12
CTB frequency rounding, CTB frequency measurement, CTB frequency units

### LINKED ISSUE #1390: CTB clocking
We should round to the next-closest possible frequency instead of rounding up. & something strange is happening for 1MHz:

Configured runclk Frequency, actual frequency (MHz):

0 21.052630
1 2.777777
2 1.999999
3 3.007518
4 3.999999
5 4.999999
6 6.015036
7 7.017543
8 7.999999
9 9.090908
10 9.999999
11 11.111110
12 12.121212
13 13.114753
14 14.035087
15 15.094338
16 15.999999
17 17.021275
18 18.181817
19 19.047618
20 19.999999
21 21.052631
22 22.222221
23 23.529410
24 24.242423
25 24.999999
26 26.666665
27 27.586205
28 28.571428
29 29.629629
30 30.769229
30 30.769230
31 31.999999
32 31.999999
33 33.333332
34 34.782607
35 36.363635
36 36.363636
37 38.095238
38 38.095237
39 39.999999
40 39.999999
41 42.105262
42 42.105263
43 44.444444
44 44.444444
45 47.058822
46 47.058822
47 47.058822
48 49.999999
49 49.999999
50 49.999999
51 53.333332
52 53.333332
53 53.333332
54 57.142856
55 57.142856
56 57.142856
57 57.142856
58 61.538460
59 61.538461
60 61.538461
61 61.538460
62 66.666666
63 66.666665
64 66.666665
65 66.666665
66 66.666666
67 72.727271
68 72.727271
69 72.727272
70 72.727272
71 72.727272
72 72.727272
73 79.999999
74 79.999999
75 79.999999
76 79.999999
77 79.999999
78 79.999999
79 79.999999
80 79.999999
81 88.888888
82 88.888888
...
87 88.888888
88 88.888888
89 99.999999
90 99.999999
...
99 99.999999
100 99.999999
101 114.285713
102 114.285713
...
113 114.285714
114 114.285714
115 133.333332
116 133.333332
117 133.333333
...
132 133.333332
133 133.333332
134 159.999999
135 159.999999
136 159.999999
...
159 159.999999
160 159.999999
161 199.999999
162 199.999999
...
198 199.999999
199 199.999999
200 199.999999
201 266.666666
202 266.666665

...

### LINKED ISSUE #1422: Max/Min Limit for CTB & XCTB clocks
There are currently not the correct checks on the CTB & XCTB if a frequency requested by "p runclk" etc. can actually be delivered by the firmware. 

The XCTB should have a lower limit of 10 MHz, CTB 2 MHz. Upper limits is something we need to look at. 

There is also a difference between "frequency can be delivered" and "firmware is guaranteed to work at this frequency". Right now the limit where the CTB firmware is expected to work is only 80 MHz for the runclk, 40 MHz for adc & syncclk. I have seen configs with higher values and i think there should at least be some kind of warning if values above that are inserted. 

I have to try if i can increase the limits, right now CTB firmware might not work properly in these cases.

mmarti04:
- Changed how frequencies of CTB & XCTB are used in the software (Hz instead of MHz)
- runclk now takes units and floating point values, no unit defaults to MHz:
```
p runclk 3
p runclk 3 MHz
p runclk 3000 kHz
p runclk 3000000 Hz
p runclk 3.5 MHz
```

- We round to the next closest divider settings.  (p runclk 181 will create a 180 MHz clock instead of a 200 MHz clock)
- Frequency is measured and can be read from the register PLL_FREQ_MEASURE_REG.
- Reading the runclk returns this measurement in Hz and not the requested frequency:
```
[muelle_m1@pc16956 bin]$ p runclk 3.5 MHz
runclk 3.5MHz
[muelle_m1@pc16956 bin]$ g runclk
runclk 3493507
```
- This introduced a delay after a frequency change since i need time in the firmware to make an accurate measurement. (p runclk blocks for 2s)
- older firmware versions return 0 when accessing PLL_FREQ_MEASURE_REG. In this case we return the "software frequency" instead of the measured one. --> should be backwards compatible

- Cycles for exposuretime etc. are calculated from the actual frequency, not the requested one:
```
[muelle_m1@pc16956 bin]$ p runclk 85.5 MHz
runclk 85.5MHz
[muelle_m1@pc16956 bin]$ g runclk
runclk 88890367
[muelle_m1@pc16956 bin]$ p patwaittime 0 1s
[muelle_m1@pc16956 bin]$ g patwaittime 0
patwaittime 0 88890367
```

- reading exptime etc. might not return the same value that was written since it returns the actual exptime that can be produced at the configured runclk:
```
[muelle_m1@pc16956 bin]$ p exptime 3us
exptime 3us
[muelle_m1@pc16956 bin]$ g exptime
exptime 2.862us
```
- pyctbgui displays and accepts runclk, dbitclk & adcclk as floating point values
- d.runclk etc. still uses MHz to keep things compatible with older configs & scripts. But it accepts decimal values:
```
>>> d.runclk=3.7548168
>>> d.runclk
3.755932
>>> 
```

TODO: change Version API and upload compiled server once PR is approved

Should we recompute patwaittimes whenever the runclk is changed ? This is also not done in the current developer branch but might be useful and not so hard to do.


mmarti04:
@erikfrojdh How do we make the behaviour for CTB & XCTB the same ? In principle we can set low frequencies much more precise than full MHz values. The measured frequency should be accurate to a few hundred Hz. Do we just round everything to the next full MHz on both boards ?

mmarti04:
Tested it on Xilinx and altera CTB, made a new release of the CTB firmware which supports frequency measurements:
/psi.ch/group/detector/firmware/ctb_firmware/cyclone_V/CTB_2067c4a9.pof
Xilinx already supports them.

I think it should not affect other detectors but i could have missed something. Please have a look when you have time. 

Will break Matterhorn configs & scripts on Xilinx which use kHz values right now but if we want the same behaviour on both CTB's i guess there is no other way.

thattil:
> Tested it on Xilinx and altera CTB, made a new release of the CTB firmware which supports frequency measurements: /psi.ch/group/detector/firmware/ctb_firmware/cyclone_V/CTB_2067c4a9.pof Xilinx already supports them.
> 
> I think it should not affect other detectors but i could have missed something. Please have a look when you have time.
> 
> Will break Matterhorn configs & scripts on Xilinx which use kHz values right now but if we want the same behaviour on both CTB's i guess there is no other way.

Suggestion at the requirements: If no unit and its taken as MHz for Xilinx, it should just throw from the server side for anything above 300MHz to inform the user to update the unit.

thattil:
As discussed with @mmarti04 default unit will be Hz. Checks in server for maximum is 300 MHz, min for ctb: 2 MHz and xilinx is 10 MHz. Min adc clk for ctb is 65MHz

slsDetectorServers/xilinx_ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1055,9 +1055,11 @@ int getNumTransceiverSamples() {
 int setExpTime(int64_t val) {
     setPatternWaitInterval(0, val);
 
-    // validate for tolerance
+    // Tolerance: three clock periods in ns.
     int64_t retval = getExpTime();
-    if (val != retval) {
+    int64_t toleranceNs = 3 * (1000000000 / clkFrequency[RUN_CLK]);
```
thattil: Is this check needed only for exptime? Also for period?

slsDetectorServers/slsDetectorServer/src/XILINX_PLL.c
```diff
@@ -11,9 +11,9 @@
 // leave some things away)
```
thattil, line 11: why is this file in the common dir?

slsDetectorServers/xilinx_ctbDetectorServer/slsDetectorFunctionList.c
mmarti04: It should probably also apply to period but i overlooked it. During testing i only ran into the exptime return FAIL case without the tolerance check. 

Do you add this in your ctb_clocks_fix PR or should i change it here ?

slsDetectorServers/slsDetectorServer/src/XILINX_PLL.c
mmarti04: i don't know. ALTERA_PLL.c was in this directory so i also put XILINX_PLL.c here.

slsDetectorServers/xilinx_ctbDetectorServer/slsDetectorFunctionList.c
thattil: We both finally decided that we will remove the 3 clock cycle tolerance and remove the +0.5 rounding on the set only and that 1 clock cycle difference is not so bad.

mmarti04:
i started a round of testing in hardware and for some reason the removal of the +0.5 rounding here https://github.com/slsdetectorgroup/slsDetectorPackage/pull/1434/changes/4de61dfb7f4d24ba0ca1f329e69fb6504a194f6f does not work.

```
[muelle_m1@pc16956 altera_ctb]$ p exptime 10000ns
- 11:56:40.407 WARNING: Exposure time is deprecated and will be removed for this detector. Please migrate to patwaittime.
- 11:56:40.410 ERROR: Detector returned: Could not set exposure time. Set 10000 ns, read 9999 ns.
```
I don't understand why right now, discussed it yesterday with @thattil

mmarti04:
OK, in this example this happens, runclk is 80.0014 MHz:

```
set: 10000ns × 0.0800014MHz = 800.014 clocks --> set num Clocks in firmware to 800
validate: read numClocks 800, 800 / 0.0800014 = 9999.825 ns --> truncate and return 9999ns --> mismatch
```

@thattil what we discussed yesterday does not make sense. 
If we convert from time to clock cycles and back then we can always end up at a different time in ns. because we cannot represent every time step with a unique number of clock cycles --> more than 1 "time in ns" will convert to a specific "N clock cycles" which will convert back to only 1 "time in ns". So there is always the possibility of the mismatch and we need the tolerance check

thattil:
> OK, in this example this happens, runclk is 80.0014 MHz:
> 
> ```
> set: 10000ns × 0.0800014MHz = 800.014 clocks --> set num Clocks in firmware to 800
> validate: read numClocks 800, 800 / 0.0800014 = 9999.825 ns --> truncate and return 9999ns --> mismatch
> ```
> 
> @thattil what we discussed yesterday does not make sense. If we convert from time to clock cycles and back then we can always end up at a different time in ns. because we cannot represent every time step with a unique number of clock cycles --> more than 1 "time in ns" will convert to a specific "N clock cycles" which will convert back to only 1 "time in ns". So there is always the possibility of the mismatch and we need the tolerance check

Yes I agree.

mmarti04:
> > OK, in this example this happens, runclk is 80.0014 MHz:
> > ```
> > set: 10000ns × 0.0800014MHz = 800.014 clocks --> set num Clocks in firmware to 800
> > validate: read numClocks 800, 800 / 0.0800014 = 9999.825 ns --> truncate and return 9999ns --> mismatch
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @thattil what we discussed yesterday does not make sense. If we convert from time to clock cycles and back then we can always end up at a different time in ns. because we cannot represent every time step with a unique number of clock cycles --> more than 1 "time in ns" will convert to a specific "N clock cycles" which will convert back to only 1 "time in ns". So there is always the possibility of the mismatch and we need the tolerance check
> 
> Yes I agree.

i added it again

python/tests/test_freq.py
```diff
@@ -0,0 +1,48 @@
...
+    f = MHz(0.5)
+    assert f.value == 500_000
+    f = kHz(2.5)
+    assert f.value == 2500
+    f = kHz(5000)
+    assert f.value == 5_000_000
+
+def test_rounding_exact():
+    f = MHz(1.234)
+    assert f.value == round(1.234 * 1_000_000)
```
AliceMazzoleni99: is it not better to test assert f.value == 1234000 to better test the indetend behavior?

python/tests/test_det_api.py
```diff
@@ -396,6 +396,193 @@ def test_patternstart(session_simulator, request):
...
+    """ Test using runclk for ctb and xilinx_ctb."""
+    det_type, num_interfaces, num_mods, d = session_simulator
+    assert d is not None
+
+    from slsdet import Hz, MHz, kHz
+
+    if det_type in ['ctb', 'xilinx_ctb']:
+        prev_runclk = d.getRUNClock()
+
+        d.runclk
```
AliceMazzoleni99, line 411: mmh should be prev_runclk = d.runclk ?

thattil: in both the command line, we get the previous value using detector API, then run all the tests.. and set it back using API. I did the same here as well for consistency.

python/tests/test_det_api.py
```diff
@@ -396,6 +396,193 @@ def test_patternstart(session_simulator, request):
...
+    """ Test using adcclk for ctb and xilinx_ctb."""
+    det_type, num_interfaces, num_mods, d = session_simulator
+    assert d is not None
+
+    from slsdet import Hz, MHz, kHz
+
+    if det_type in ['ctb', 'xilinx_ctb']:
+        prev_adcclk = d.getADCClock()
+
+        d.adcclk
```
AliceMazzoleni99, line 467: again remove or use to get pre_adcclk

mmarti04:
@thattil some more testing .. i ran into the same problem again for patwaittime:
```
[muelle_m1@pc16956 altera_ctb]$ p patwaittime 0 1us
- 14:32:04.714 ERROR: Detector returned: Could not set pattern Loop 0 wait time. Set 1000, but read 1130
```
i added now also the tolerance check for that in [bd77ea2](https://github.com/slsdetectorgroup/slsDetectorPackage/pull/1423/commits/bd77ea231b568ab7342e8cd67e0c7f4487a040b7) but i think that we are checking once to often here no ? Both validates check the same thing.

python/tests/test_det_api.py
```diff
@@ -450,7 +637,7 @@ def test_v_limit(session_simulator, request):
     Log(LogLevel.INFOGREEN, f"✅ {request.node.name} passed")
 
 
-@pytest.mark.detectorintegration
```
AliceMazzoleni99, line 453: why has the marker been removed?

slsSupportLib/include/sls/ToString.h
```diff
@@ -108,6 +108,44 @@ ToString(From t) {
     }
 }
 
+/** Convert frequency with specified output unit */
+template <typename T, typename Rep = double>
+typename std::enable_if<is_frequency<T>::value, std::string>::type
+ToString(T f, const std::string &unit) {
+    double val = static_cast<double>(f.value);
+
+    auto unitLower = [&] {
```
AliceMazzoleni99: I would rather add a enum Frequency with MHz, kHz instead of passing a string and checking upper lower case etc.

slsSupportLib/include/sls/ToString.h
```diff
@@ -108,6 +108,44 @@ ToString(From t) {
     }
 }
 
+/** Convert frequency with specified output unit */
+template <typename T, typename Rep = double>
```
AliceMazzoleni99: Rep is not needed

slsSupportLib/include/sls/ToString.h
```diff
@@ -108,6 +108,44 @@ ToString(From t) {
...
+typename std::enable_if<is_frequency<T>::value, std::string>::type
```
AliceMazzoleni99: are the templates neccessary and the type traits. Cant we just have an overload with Hz?

pyctbgui/pyctbgui/services/Acquisition.py
```diff
@@ -195,20 +192,49 @@ def setReadOut(self):
...
-    def getRunFrequency(self):
-        self.view.spinBoxRunF.editingFinished.disconnect()
-        self.view.spinBoxRunF.setValue(self.det.runclk)
-        self.view.spinBoxRunF.editingFinished.connect(self.setRunFrequency)
 
-    def setRunFrequency(self):
-        self.view.spinBoxRunF.editingFinished.disconnect()
+    def _getFrequency(self, det_attr, spinbox, combobox, setter):
+        spinbox.editingFinished.disconnect()
+        combobox.currentIndexChanged.disconnect()
```
mmarti04: i like the idea of the Drop-down for MHz, kHz, Hz .. but it does not work right now because of the frequency limits.

One can never switch away from "MHz"
```
Detector returned: Cannot set frequency to 0.003540 MHz. Frequency outside limits (2.000000 - 300.000000 MHz)
```
i guess we would need to multiply/divide the value by 1000 when the drop-down is changed.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -66,7 +67,8 @@ uint8_t adcEnableMask_10g = 0xFF;
 uint32_t transceiverMask = DEFAULT_TRANSCEIVER_MASK;
 
 int32_t clkPhase[NUM_CLOCKS] = {};
-uint32_t clkFrequency[NUM_CLOCKS] = {40, 20, 20, 200};
+uint32_t clkFrequency[NUM_CLOCKS] = {DEFAULT_RUN_CLK, DEFAULT_ADC_CLK,
```
AliceMazzoleni99: have the previous values been wrong ?

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -2218,13 +2219,12 @@ int getMaxPhase(enum CLKINDEX ind) {
...
     }
-    int ret = ((double)PLL_VCO_FREQ_MHZ / (double)clkFrequency[ind]) *
+    int ret = ((double)PLL_VCO_FREQ_HZ / (double)clkFrequency[ind]) *
               MAX_PHASE_SHIFTS_STEPS;
 
     char *clock_names[] = {CLK_NAMES};
-    LOG(logDEBUG1,
-        ("Max Phase Shift (%s): %d (Clock: %d MHz, VCO:%d MHz)\n",
-         clock_names[ind], ret, clkFrequency[ind], PLL_VCO_FREQ_MHZ));
+    LOG(logDEBUG1, ("Max Phase Shift (%s): %d (Clock: %d MHz, VCO:%d Hz)\n",
```
AliceMazzoleni99: not Mhz Hz

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
mmarti04: hardcoded and wrong, yes. The firmware powers up into 80,40,40,200. Maybe that was changed on the firmware end at some point but i don't remember touching this.

thattil: @mmarti04 but are the powers at start up at their max values?

mmarti04: I looked it up, it was changed 7 years ago in the firmware here: https://gitea.psi.ch/detectors_firmware/altera_hdl_library/commit/b4b6fca4445cd2875980bc13057ff4b5eaf50910

@thattil : kind of .. it is the max. frequency that the firmware was simulated for. If we have the worst FPGA Intel has ever produced and run it at 100 °C then it would be the maximum. In reality we can go higher and are not sure where the actual limit is. I'm working on getting it validated for higher frequencies but i have some trouble with that.

python/tests/test_det_api.py
AliceMazzoleni99: ah but is d.runclk just a test that command works?

thattil: yes. that it doesn throw

python/tests/test_det_api.py
thattil: I didnt get you here.. remove what or use what?

python/tests/test_det_api.py
thattil: merging issues. It showing a bit messed up here. but if you look at the file, it looks fine.

slsSupportLib/include/sls/ToString.h
thattil: I need the templates for StringTo

pyctbgui/pyctbgui/services/Acquisition.py
thattil: can one not do 1500 kHz? ill check

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: As far as I understand, the default hardcoded values on the detector server software for the clocks are the ones that the detector will start out with and not necessarily the maximum that it can go to. So whenever the default values were changed in the past, they have been communicated and have been changed. But its possible that ctb server hasnt changed its default clocks in years as they just change it in the config file by default.

Maybe my question is: are these maximum values what the firmware on the detector starts out with at power on?

python/tests/test_det_api.py
AliceMazzoleni99: same as on top. Can stay.

slsSupportLib/include/sls/ToString.h
AliceMazzoleni99: yes but not for ToString or am i misunderstanding something?

pyctbgui/pyctbgui/services/Acquisition.py
mmarti04: the highest one can put in is 300 MHz and from there one cannot switch to kHz because 300 kHz is out of range

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
mmarti04: When you give power to the firmware it will have the values 80,40,40,200 MHz, even before the software is running. Then the software can change these values by reconfiguring the Altera_pll. The PLL can produce frequencies between 800 MHz and 1.5625 MHz.

However, at 800 MHz none of the things that are connected to it will work (adc's, pattern SM, etc.) and it is unclear where between 800 MHz and 80,40,40,200 MHz things stop to work. Things are currently validated/simulated for 80,40,40,200 MHz.

The 300 MHz upper limit that i told you last week is more of a gut feeling.

python/tests/test_freq.py
thattil: done :+1:

slsSupportLib/include/sls/ToString.h
thattil: No, you are right. I tried to implement it like 'time', but there there could be many types with chrono and the ToString also required template. Thanks, that is much simpler for ToString with frequency. Ill keep the templates and type traits just for StringTo.

slsSupportLib/include/sls/ToString.h
thattil: Yes this makes more sense. At some points, maybe we should do the same with the timer equivalent. I have to change the python codegen script a bit to make it work for enum, but will do this.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: ah thanks for the explanation. I now understand that the values you mentioned are the default and the max is something else.

I realize there is some info missing why this worked for 7 years. This hardcoded values were as good as writing 0 because they are set up again to default values in the the detector set up. I will initialize them at 0 now and leave the default values as is at detector set up.

slsSupportLib/include/sls/ToString.h
thattil: Done!

slsSupportLib/include/sls/ToString.h
thattil: Done :+1:

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
thattil: Ctb: I have changed the initialization values to 0 and set up the default value at detector setup, just like xilinx.

pyctbgui/pyctbgui/services/Acquisition.py
thattil: I had a look. Alice suggested to change the value when you change the unit. That way you can change the unit and still be able to set the value again afterwards. It introduces an order but that seems the only way to set it since we technically only have MHz range.

Not doing anything when setting the unit also seems counter intuitive. So, changing unit will get the value in that unit and display to that value

thattil: Alright. this is done. Changing unit updates value in that unit. Refresh always gets value in the chosen unit. The default unit is in MHz.

thattil:
Final Summary of Changes (also picked up from @mmarti04 's description):

These changes affect 
- only Ctb and Xilinx CTb and M3 (patternwaitInterval got wrong clock)
- directly affects runclk, dbitclk, adcclk, 
- indirectly affects exptime, period, delay, patwaittime (based on runclk or sync clk), periodl, delayl

Command line:
```
p runclk 3 # no unit default in Hz
p runclk 3 MHz
p runclk 3000 kHz
p runclk 3000000 Hz
p runclk 3.5 MHz
g runclk MHz
g runclk #default in Hz

# Similarly, for adcclk, dbitclk
```
python
```
>>> from slsdet import MHz, kHz, Hz # you usually need only one
>>> d.runclk
>>> 10MHz
>>> d.runclk = MHz(5)
>>> d.runclk = Hz(5 * 1000 * 1000)
>>> d.runclk = kHz(2000)
>>>d.runclk.value = 2000
>>> c = MHz(2)
>>> for rc in [5, 10, 15, 20]:
>>>     d.runclk = rc * c
>>> assert MHz(1) == kHz(1000)
```
Detector API:
```
Strong type defs::Hz
eg. void setADCClock(defs::Hz val, Positions pos = {});
val.value =>int
```
pyctbgui:
- the units are drop down list and by default in MHz
- When unit changed, values are retrieved for the chosen unit. Similarly, when changing tabs
- Changing spinbox values sets the frequencies.



**Min, Max, rounding, measured:**
- 300MHz is max, ctb min: 2MHz, xilinx min: 10MHz, max adc clk: 65MHz (checked at server level).
- Frequency is measured and can be read from the register PLL_FREQ_MEASURE_REG. More accurate now: 80MHz=> 80.014MHz, but requires time (~2s). Also retrieved when setting frequency.
    - Effects of accuracy:
```
$ p runclk 85.5 MHz
runclk 85.5MHz
$ g runclk
runclk 88890367
$ p patwaittime 0 1s
$ g patwaittime 0
patwaittime 0 88890367
```
```
$ p exptime 3us
exptime 3us
$ g exptime
exptime 2.862us
```
- Round to the next closest divider settings. (p runclk 181 will create a 180 MHz clock instead of a 200 MHz clock). Tolerance check is 1 clock cycle.
- older firmware versions return 0 when accessing PLL_FREQ_MEASURE_REG. In this case we return the "software frequency" instead of the measured one. --> should be backwards compatible (?)

**patwaittime recompute?**
Was decided not to recompute this as it could have been set in cycles instead of intervals. Similarly, period and delay are also not updated when changing frequency as before. User should set these values if not zero after changing frequency in config file.

**Breaking Ctb and Xilinx Ctb configs and scripts**
Was not possible to keep it clean and intuitive with units without breaking compatibility with old config styles. When using old config values, which assume MHz or kHz, it will be misinterpreted as Hz and instead of silently accepting these values, the checks in the server will inform you of the min and max values.

**conversion from timer to frequency**
- conversion in common.h 
- checking for division by 0 ( if clk is 0) => changed signature of set/get of period, exptime, delayaftertrigger for Ctb and xilinx ctb
- validate64_timer validates the conversion with 1 clock cycle tolerance

** actual time and measurement time **
- were not connected in firmware, now changed to reg contents in units of 100ns

**M3 Patternwaittime**
It was calculating the patwaittime incorrectly by multiplying with clock divider and not frequency.

**Xilnix: virtual stop server commands**
The virtual stop server did not set up detector, and so frequency was at 0. Commands using them such as delayl and periodl that go to stop server gave weird values because of that. Fixed it to setupDetector for virtual stop servers (just like other detectors)

** Tests **
Tests also include testing timers and accuracy when setting runclk etc.


Others:
- Added pyzmq for workflow tests
- Detector API: fixed getPatternFileName typo

python/tests/test_det_api.py
thattil: it was missing from merging, but i kept missing it. added it now.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1134,64 +1141,93 @@ int setNumTransceiverSamples(int val) {
...
-    int64_t retval = getPeriod();
-    val /= (1E-3 * clkFrequency[SYNC_CLK]);
-    if (val != retval) {
+    int64_t retval = 0;
+    int ret = getPeriod(&retval, mess);
+    if (ret == FAIL) {
         return FAIL;
     }
-    return OK;
+    validate64_timer(&ret, mess, val, retval, clkFrequency[SYNC_CLK], "period");
```
AliceMazzoleni99: Instead of validating periods check if(numClocks == get64BitReg(PERIOD_LSB_REG, PERIOD_MSB_REG)). I think this is cleaner and does not involve any rounding checks. However maybe add a LOG(logINFO, (Rounding to %llds ns\n", (long long int)getPeriod(6retval, mess))

AliceMazzoleni99 (approved):
Small comment. Otherwise looks good to me.

PR #1424, opened 2026-03-17
dev ctb, xilinx: safe power on

thattil:
- ctb and xilinx: setting all dacs (normal, not power dacs) to 0 (not power down) at startup. This is safer than power down for 4 normal dacs. 

- xilinx ctb: remove disable fmc at power off chip so one can power on and off the chip on their own


AliceMazzoleni99 (approved):
Looks good to me.

PR #1427, opened 2026-03-31
Running Matterhorn on altera CTB

mmarti04:
readout of Matterhorn on altera CTB via SPI works, ~5s per image. 

Bit-selection of mosi, sclk, etc. is still hardwired in the firmware. Once that is selectable i would add the bit-selection via p define in the config (similar to the XCTB) --> We don't have to touch the software again, SW changes should be complete with this PR.


mmarti04:
Bit selection now also works, same as XCTB (https://gitea.psi.ch/detectors/chiptesting/commit/744ef8e591a2c5441fdab944cf9a32c0446d2605)

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11064,14 +11064,23 @@ int set_pattern_wait_interval(int file_des) {
     return Server_SendResult(file_des, INT64, NULL, 0);
 }
 
+// usleep is not viable on blackfin
+void usleep_bf(uint64_t i){
```
AliceMazzoleni99, line 10911: ah great make sure to also use usleep_bf for moench, jungfrau for other detectors. e.g. in initial_checks

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11211,6 +11220,15 @@ int spi_read(int file_des) {
         sprintf(mess, "SPI write failed with %d:%s\n", errno, strerror(errno));
         return sendError(file_des);
     }
+#elif defined(CHIPTESTBOARDD)
+    // set spi to 8 bit per word (-1 comes from the firmware), set chipselect
+    bus_w(SPI_CTRL_REG, ((8 - 1) << SPI_CTRL_NBIT_OFST)+ (1 << SPI_CTRL_CHIPSELECT_BIT));
+    for (int i = 0; i < n_bytes + 1; ++i) {
+        // TODO: should we make bus_w to this address blocking in the firmware to remove usleep ?
+        bus_w(SPI_WRITEDATA_REG, local_tx[i]);
+        usleep_bf(25);
```
AliceMazzoleni99: dont hardcode 25

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
```diff
@@ -11245,6 +11263,13 @@ int spi_read(int file_des) {
         return sendError(file_des);
     }
     close(spifd);
+#elif defined(CHIPTESTBOARDD)
+    for (int i = 0; i < n_bytes + 1; ++i) {
+        bus_w(SPI_WRITEDATA_REG, local_tx[i]);
+        usleep_bf(25);
+        local_rx[i] = (uint8_t)bus_r(SPI_READDATA_REG);
```
AliceMazzoleni99, line 11113: Is this read neccessary?

mmarti04: in principle not but SPI_READDATA_REG is not a simple register but the read-port of a FIFO-buffer. So whenever we write with bus_w(SPI_WRITEDATA_REG, local_tx[i]); there is one word added to the buffer at SPI_READDATA_REG and these words stay there until we read them. 

If we don't read them here i have to add a way to clear this buffer somehow, otherwise the next read would be wrong.

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
mmarti04: you mean i should make this dependent on the configured clock-speed ? 
Or define a constant somewhere ?

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
mmarti04: in inital_checks it is a 3s usleep, i think doing the "cpu-busy" usleep_bf there does not make sense for 3s. Only if we need really something on the scale of micro-seconds i would do usleep_bf()

slsDetectorServers/slsDetectorServer/src/slsDetectorServer_funcs.c
AliceMazzoleni99: define a constant

mmarti04: done

AliceMazzoleni99 (approved):

PR #1428, opened 2026-04-08
CTB: simplify power monitoring

mmarti04:
Implementation to read power and current was waaaaaay too complicated in the firmware and also software.
Now we can just read power & current directly with bus_r()

probably in conflict with PR #1420 but should not be a big problem since i am mostly deleting stuff.


slsDetectorServers/ctbDetectorServer/RegisterDefs.h
```diff
@@ -636,6 +636,9 @@
 #define SPI_WRITEDATA_REG (0x49 << MEM_MAP_SHIFT)
 #define SPI_READDATA_REG (0x4A << MEM_MAP_SHIFT)
 
+/* Power monitor base register */
+#define POWER_MONITOR_BASE_REG (0x50 << MEM_MAP_SHIFT)
+
 /** I2C Control register */
 #define I2C_TRANSFER_COMMAND_FIFO_REG (0x100 << MEM_MAP_SHIFT)
```
thattil: These register definitions should be removed I believe?

mmarti04: yes

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1767,26 +1755,35 @@ int getPowerADC(enum powerIndex index, int *retval, char *mess) {
...
-    case I_POWER_D:
         adcIndex = V_PWR_D;
         break;
     case V_POWER_IO:
-    case I_POWER_IO:
         adcIndex = V_PWR_IO;
         break;
-
+    case I_POWER_A:
+        adcIndex = I_PWR_A;
```
AliceMazzoleni99, line 1773: why only  measured current not measured voltage?

AliceMazzoleni99 (commented):
Some comment. Dont forget to compile the servers.

slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
mmarti04: because before adcIndex was used to calculate the deviceId and then in line 1799 of the old file we had the check  if (index < I_POWER_A) { --> read Voltage} else {--> readCurrent}.

Now there is an index for each measurement and the concept of deviceId is gone.

PR #1435, opened 2026-04-20
dev/ fix vchip and binaries

thattil:
- set vchip every time power dac also changes (currently only for power rail on/off)
   - recalculate vchip value (get current power rail enables and values, replace with current command to enable/disable all or change power dac)
- initialize powerValues to 0 instead of -1 as the initialize power dacs values to minimum at the start requires powerValues to have reasonable values prior.
- added tests to check vchip also changes when changing power dac
- moved usleep_bf to blackfin to eliminate warning
- recompiled all binaries (hence all the detector labels)
- 

related to #1420 (bug from it)


slsDetectorServers/ctbDetectorServer/slsDetectorFunctionList.c
```diff
@@ -1585,17 +1600,28 @@ int setPowerEnabled(enum powerIndex indices[], int count, bool enable,
...
+        return FAIL;
+    // update with current command
+    for (int i = 0; i != count; ++i) {
+        pwrEnables[indices[i]] = enable;
+    }
+
     // set vchip accordingly
     int vchipToSet = 0;
-    if (getVchipToSet(&vchipToSet, mess) == FAIL)
+    if (getVchipToSet(&vchipToSet, pwrEnables, pwrValues, mess) == FAIL)
```
AliceMazzoleni99: maybe a better name would be calculate_voltage_Vchip

AliceMazzoleni99 (approved):
To me it looks good. I added a commit directly that fixes some compiler errors and some documentation mismatch.

PR #1439, opened 2026-04-22
Add support for building RPMs

### LINKED ISSUE #1234: Add support for building rpms
Having RPMs would help with deployment at PSI and generating them seems relatively easy using CMake:
https://cmake.org/cmake/help/latest/cpack_gen/rpm.html

**Multi step approach**

- Add Cpack support
- Split up Components to allow for specific RPMS

erikfrojdh:
RPMs is the easiest way to auto deploy software to consoles and servers at PSI. The idea is to build them on the standard gitea runner and push to our internal repository. But of course this can also be useful outside of PSI. 

This is a minimal PR that contains:

1. Option SLS_INSTALL_VERSIONED_BINARIES to append version number to binaries
2. Added support for RPMs using CPack
3. Separate cli, receiver and zmq components for more fine grained install. 

The options for appending version number is important since we want to be able to install multiple versions side by side. 

Ideally we backport this to at least version 9 and 10 this is why I want to have the absolute smallest change that works.


AliceMazzoleni99 (commented):
Looks good to me.  Should we also create an rpm for the GUI and the other detector processing targets?

erikfrojdh:
I can include the jungfrau zmq stuff in the zmq rpm. For the GUI I need to add qt but looks easy.

AliceMazzoleni99 (approved):
Tested the GUI and it works fine also with no pyqt5 installed beforehand.

PR #1442, opened 2026-04-30
Add rx_streamdummyheader command

### LINKED ISSUE #1095: zmq streaming: Change the dummy streaming method
change the dummy streaming method or at least put in the changes from this issue #1081
@anberga

lrlunin:
In this PR I implement the `rx_restream_stop` put command.

This allows to send a ZMQ dummy header any time user wants to do so. For example this allows to pre-configure the ZMQ processing software before the acquisition begins. Therefore, the dummy header was adapted in order to contain the fields stored in the receiver.

**Known issues:**
1. `fname` field is not properly set before the first acquisition:

This happes because the receiver variable for the filename - `fileNametoStream` - is assigned in the `void DataStreamer::ResetParametersforNewAcquisition(const std::string &fname)` function call, which is called in the `rx_start` (or in `rx_stop` if the receiver was not in idle, which implies that the `rx_start` was called before).  So, the ZMQ dummy header has the corrent `fname` data after first `rx_start -> rx_stop` sequence.

P.S. python bindings are missing too, but I'll add them.


slsDetectorSoftware/generator/commands.yaml
```diff
@@ -1636,6 +1636,13 @@ rx_stop:
     PUT:
       function: stopReceiver
 
+rx_restream_stop:
```
thattil: I would stick to "rx_restreamstop". We try to use less underscores in the naming.

slsDetectorSoftware/generator/commands.yaml
```diff
@@ -1636,6 +1636,13 @@ rx_stop:
...
+  inherit_actions: EXECUTE_SET_COMMAND_NOID
```
thattil: I would rather use EXECUTE_SET_COMMAND, so that one can individually send this command to a single module as well. Accordingly, the signatures will change. Info: NOID=>Requires no ID (for all modules)

slsDetectorSoftware/generator/commands.yaml
```diff
@@ -1636,6 +1636,13 @@ rx_stop:
...
+  help: "\n\tSends ZMQ dummy header. The detector related fields are not set, the receiver related fields are set."
+  actions:
+    PUT:
+      function: restreamStop
+
```
thattil, line 1707: A general comment: could it be that the autocompletion script generator python script was not run? I  see that the slsDetectorSoftware/generator/autocomplete/bash_autocomplete.sh (and zsh_autocomplete.sh) scritps dont seem to have any changes. 

cd slsDetectorSoftware/generator
python gen_commands -agf

slsDetectorSoftware/include/sls/Detector.h
```diff
@@ -700,6 +700,8 @@ class Detector {
     /** Gets Scan error message if scan ended in error for non blocking
      * acquisitions.*/
     Result<std::string> getScanErrorMessage(Positions pos = {}) const;
+
+    void restreamStop();
```
thattil: 1. I would move this to anywhere under the zmq section as discussed before as its  important only there. 
2. I would also rename this function with and 'Rx' in it so that its clear it has to do with the Receiver. 'restreamRxStop?' or something similar?
3. I would include the positions. =>void restreamRxStop(Positions pos = {});

slsReceiverSoftware/src/DataStreamer.cpp
```diff
@@ -177,24 +195,19 @@ int DataStreamer::SendDataHeader(sls_detector_header header, uint32_t size,
     zHeader.data = true;
     zHeader.jsonversion = SLS_DETECTOR_JSON_HEADER_VERSION;
 
+    // parameter coming from the detector
```
thattil: The only parameters coming from the detector header are the ones from framenumber to version that is now not in order. I would these header info back how it was in the right order and then put this comment to it.

thattil: I would change this to "parameters from the detector (raw and derived)".. because some dont come from the detector and are instead derived from the receiver:
* progress: is of the receiver processing frame
* packetnumber: packets caught by receiver
* completeImage: if reciever udp port has caught the complete image.

slsReceiverSoftware/src/DataStreamer.cpp
```diff
@@ -168,6 +168,24 @@ int DataStreamer::SendDummyHeader() {
...
     zHeader.data = false;
     zHeader.jsonversion = SLS_DETECTOR_JSON_HEADER_VERSION;
+
+    // parameters coming from the receiver software
+    zHeader.fname = fileNametoStream;
+    zHeader.dynamicRange = generalData->dynamicRange;
+    zHeader.fileIndex = fileIndex;
+    zHeader.ndetx = numPorts.x;
+    zHeader.ndety = numPorts.y;
+    zHeader.flipRows = static_cast<int>(flipRows);
```
thattil: Could you include the following as well, just like in the normal header?

npixelsx = generalData->nPIxelsX; 
npixelsy = generalData->nPixelsY; 
imageSize = generalData->imageSize;
if (generalData->detType == GOTTHARD2 && index ! = 0) {
       imageSize = generalData->vetoImageSize;
}

thattil (changes requested):
Looks good overall. I have requested a few changes.

As for the FilenametoStream,
I would edit the Implementation.cpp::restreamStop to add the argument as its visible here.

std::string fnametostream = (filePath / fileName).string();
        for (const auto &it : dataStreamer)
            it->RestreamStop(fnametostream);

So, I would just add an argument here with the file name to stream and it should resolve the issue. One do not need to do rx_Start and rx_stop for this.

slsReceiverSoftware/src/DataStreamer.cpp
lrlunin: Should I also change this? Because originally the `ndetx`/`ndety` were set from the arguments of 

```cxx
int DataStreamer::SendDataHeader(sls_detector_header header, uint32_t size,
                                 uint32_t nx, uint32_t ny)
```

https://github.com/slsdetectorgroup/slsDetectorPackage/blob/4ffb81e7ffa02ac7c92eef2135ff8137f7ef5745/slsReceiverSoftware/src/DataStreamer.cpp#L187-L188

lrlunin (commented):
Previous comments are resolved. The python binding are missing. Can you add them when PR is ready?

slsDetectorSoftware/generator/commands.yaml
thattil: As discussed,
New suggestion for command: rx_streamdummyheader 
Detector API: streamRxDummyHeader
@AliceMazzoleni99

slsDetectorSoftware/generator/commands.yaml
```diff
@@ -1636,6 +1636,13 @@ rx_stop:
     PUT:
       function: stopReceiver
 
+rx_restreamstop:
```
thattil: As discussed,
New suggestion for command: rx_streamdummyheader 
Detector API: streamRxDummyHeader
@AliceMazzoleni99

slsReceiverSoftware/src/DataStreamer.cpp
```diff
@@ -152,8 +151,7 @@ void DataStreamer::ProcessAnImage(sls_detector_header header, size_t size,
     uint64_t fnum = header.frameNumber;
     LOG(logDEBUG1) << "DataStreamer " << index << ": fnum:" << fnum;
 
-    if (!SendDataHeader(header, size, generalData->nPixelsX,
-                        generalData->nPixelsY)) {
+    if (!SendDataHeader(header, size)) {
```
thattil, line 157: thank you! 👍

slsReceiverSoftware/src/DataStreamer.cpp
```diff
@@ -218,11 +189,49 @@ int DataStreamer::SendDataHeader(sls_detector_header header, uint32_t size,
...
+}
+
+int DataStreamer::SendDataHeader(sls_detector_header header, uint32_t size) {
+    zmqHeader zHeader = prepareRxZmqHeader();
+    zHeader.data = true;
+
+    // parameter coming from the detector (raw and derived)
+    uint64_t frameIndex = header.frameNumber - firstIndex;
+    uint64_t acquisitionIndex = header.frameNumber;
+
```
thattil: Im sorry for nitpicking, but could you please maintain this order so its easier to see whats included?


// det header
zHeader.frameNumber 
zHeader.expLength
zHeader.packetNumber
zHeader.detSpec1
zHeader.timestamp
zHeader.modId 
zHeader.row
zHeader.column 
zHeader.detSpec2
zHeader.detSpec3
zHeader.detSpec4
zHeader.detType
zHeader.version

// derived from rx
zHeader.acqIndex 
zHeader.frameIndex 
zHeader.completeImage
zHeader.progress
zHeader.imageSize

slsReceiverSoftware/src/DataStreamer.cpp
```diff
@@ -218,11 +189,49 @@ int DataStreamer::SendDataHeader(sls_detector_header header, uint32_t size,
...
+        100 * ((double)(frameIndex + 1) / (double)(nTotalFrames));
+    zHeader.row = header.row;
+    zHeader.timestamp = header.timestamp;
+    zHeader.version = header.version;
 
     return zmqSocket->SendHeader(index, zHeader);
 }
 
-void DataStreamer::RestreamStop() {
+void DataStreamer::RestreamStop(const std::string &fname) {
```
thattil: As discussed earlier, you could have a function to set the filename, which is called before calling SendDummyHeader and then you would not need this function anymore.

thattil (changes requested):
Looks good. The name change as we discussed before and 1-2 minor ones. 

Thank you!

slsReceiverSoftware/src/DataStreamer.cpp
lrlunin: Separated in two groups as suggested, in each group the parameters are sorted alphabetically.
```
    // parameter from detector header
    zHeader.column = header.column;
    zHeader.detSpec1 = header.detSpec1;
    zHeader.detSpec2 = header.detSpec2;
    zHeader.detSpec3 = header.detSpec3;
    zHeader.detSpec4 = header.detSpec4;
    zHeader.detType = header.detType;
    zHeader.expLength = header.expLength;
    zHeader.frameNumber = header.frameNumber;
    zHeader.modId = header.modId;
    zHeader.packetNumber = header.packetNumber;
    zHeader.row = header.row;
    zHeader.timestamp = header.timestamp;
    zHeader.version = header.version;

    // parameter derived from header and receiver
    uint64_t frameIndex = header.frameNumber - firstIndex;
    uint64_t acquisitionIndex = header.frameNumber;
    zHeader.acqIndex = acquisitionIndex;
    zHeader.completeImage =
        (header.packetNumber < generalData->packetsPerFrame ? false : true);
    zHeader.frameIndex = frameIndex;
    zHeader.imageSize = size;
    zHeader.progress =
        100 * ((double)(frameIndex + 1) / (double)(nTotalFrames));
```

slsReceiverSoftware/src/DataStreamer.cpp
lrlunin: Added `SetFileName` function to `DataStreamer`. Also replaced filename assignment in `ResetParametersforNewAcquisition` in favour of setting the filename with `SetFileName` after (order of assignment persists).

slsReceiverSoftware/src/DataStreamer.cpp
thattil: Could you keep the order of the detector header though instead of sorting them alphabetically? It makes reviewing or debugging easier because we can compare the mapping directly against the detector header definition and quickly spot missing or newly added fields.

thattil:
> Previous comments are resolved. The python binding are missing. Can you add them when PR is ready?

Just added python bindings. Reviewing your changes now.

slsDetectorSoftware/include/sls/Detector.h
```diff
@@ -1162,6 +1162,7 @@ class Detector {
      */
     void setRxZmqHwm(const int limit);
 
+    void restreamRxStop(Positions pos = {}) const;
```
thattil: 👍

slsDetectorSoftware/src/Detector.cpp
```diff
@@ -1636,6 +1636,10 @@ void Detector::setRxZmqHwm(const int limit) {
     }
 }
 
+void Detector::streamRxDummyHeader(Positions pos) const {
+    pimpl->Parallel(&Module::restreamStopFromReceiver, pos);
```
thattil: I would also change the function name in the Module, enum for detector_funcs.h and in the receiver side in ClientInterface, Implementation, DataStreamer for consistency.

thattil (changes requested):
Just the previous comment about keeping the detector header order + consistent function names across the package. If you prefer or dont have time, I could do this for you. Let me know :)

thattil (approved):
LGTM

PR #1445, opened 2026-05-04
ctb: bugfix of current display & fw version

### LINKED ISSUE #1443: CTB: Negative current measurement
The current measurement on the chiptestboards can return negative values. They are currently not correctly displayed:
```
(det) [muelle_m1@pc16956 altera_ctb]$ g im_a
im_a 27351 mA
```
(should be -1 mA)

mmarti04:
the measured current on the ctb can be negative. I missed that in PR #1428


mmarti04:
i was wondering why we did not have this problem in the old ina226.c that was deleted in PR #1428. 

Turns out it was also there and was "solved/ignored" by reading again in case it was negative:

```
    int cuurentRegVal = I2C_Read(deviceId, INA226_CURRENT_REG);
    LOG(logDEBUG1, (" current reg: %d\n", cuurentRegVal));
    // read it once more as this error has occured once
    if (cuurentRegVal >= 0xFFF0) {
        LOG(logDEBUG1, (" Reading current reg again\n"));
        cuurentRegVal = I2C_Read(deviceId, INA226_CURRENT_REG);
        LOG(logDEBUG1, (" current reg: %d\n", cuurentRegVal));
    }
```

thattil (approved):

PR #1446, opened 2026-05-05
Fixing error in setting dacs due to deprecated np.int

### LINKED ISSUE #1441: Setting dacs from array
Include if we do a bugfix release for v10

```bash  File "/home/l_msdetect/erik/sls_detector_tools/sls_detector_tools/calibration.py", line 75, in setup_measurement
    detector.dacs.set_from_array( dacs )
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/l_msdetect/miniforge3/envs/root-env/lib/python3.13/site-packages/slsdet/dacs.py", line 185, in set_from_array
    dac_array = dac_array.astype(np.int)
                                 ^^^^^^
  File "/home/l_msdetect/miniforge3/envs/root-env/lib/python3.13/site-packages/numpy/__init__.py", line 775, in __getattr__
    raise AttributeError(__former_attrs__[attr], name=None)
AttributeError: module 'numpy' has no attribute 'int'.
`np.int` was a deprecated alias for the builtin `int`. To avoid this error in existing code, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

erikfrojdh:
- using np.int32 instead of the deprecated np.int
- fixed dacs being return as float (default dtype of numpy)

closes #1441

### DIFF
--- a/python/slsdet/dacs.py
+++ b/python/slsdet/dacs.py
@@ -89,7 +89,7 @@ class NamedDacs:
         """
         Read the dacs into a numpy array with dimensions [ndacs, nmodules]
         """
-        dac_array = np.zeros((len(self._dacnames), len(self._detector)))
+        dac_array = np.zeros((len(self._dacnames), len(self._detector)), dtype = np.int32)
         for i, _d in enumerate(self):
             dac_array[i,:] = _d[:]
         return dac_array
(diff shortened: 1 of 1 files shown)


AliceMazzoleni99 (approved):
Looks good.

PR #1447, opened 2026-05-06
Wrapping binaries in python pkg

erikfrojdh:
Prototype implementation to wrap sls_detector_get/put/acquire + slsReceiver for distribution with pip. 

Similar implementation to the cmake package from scikit-build: https://github.com/scikit-build/cmake-python-distributions/tree/main using os.execl to replace process. 

Pros:
- Easy install of command line and receiver without depending on conda 

Cons:
- Longer build time for wheels
- Larger wheels
- Potential conflict with installed binaries?


erikfrojdh:
Working implementation. Need to decide if we want to do this. Main concerns is how it would play with conda and that we need the receiver running on macOS

erikfrojdh:
Cherry pick and merge if we get a request.

PR #1449, opened 2026-05-08
removed legacy patternGenerator

### LINKED ISSUE #1397: Remove/move the C pattern generator
With the python pattern generator integrated in slsdet is there any reason to not remove the C version from the library?

erikfrojdh:
Removed the legacy C/Macro based pattern generator since it has been replace with the python based pattern generator in slsdet. 

Archive of the old generator: https://github.com/slsdetectorgroup/patternGenerator

closes #1397


thattil (approved):
Just a minor comment. Do you want to also remove the options in the pyctbgui to compile and load the old type (c) patternGenerator?

erikfrojdh:
Decided to keep the option since users have to pass an absolute path. Might be removed later when everyone is fully transitioned away from the old pattern generator.

PR #1463, opened 2026-05-20
accept comma separated rx_dbitlist

### LINKED ISSUE #1340: rx_dbitlist only takes first value when providing a comma separated list
p rx_dbitlist 1,2,3
rx_dbitlist [1,2,3]
[froejdh_e@pc-moench-03 ~]$ g rx_dbitlist
rx_dbitlist [1]

erikfrojdh:
_**PR included fix to github actions needed for conda build to work**_ 
[https://github.com/slsdetectorgroup/aare/pull/309](url)

**Actual change to detector software**:

- Accept comma separated list of rx_dbitlist without requiring space

```bash
p rx_dbitlist 12,13 45, 27
```

- Added same parsing for counters

Open question should we do a get, reformat args or just print what the user put in?

### DIFF
--- a/slsDetectorSoftware/src/CallerSpecial.cpp
+++ b/slsDetectorSoftware/src/CallerSpecial.cpp
@@ -1014,15 +1014,16 @@ std::string Caller::counters(int action) {
         if (args.empty()) {
             WrongNumberOfParameters(1);
         }
-        if (std::any_of(args.cbegin(), args.cend(), [](std::string s) {
-                return (StringTo<int>(s) < 0 || StringTo<int>(s) > 2);
+        //convert args to string and then to a vector of ints
+        auto counters = StringTo<std::vector<int>>( ToString(args));
+        if (std::any_of(counters.cbegin(), counters.cend(), [](int val) {
+                return (val < 0 || val > 2);
             })) {
             throw RuntimeError("Invalid counter indices list. Example: 0 1 2");
         }
         // convert vector to counter enable mask
         uint32_t mask = 0;
-        for (size_t i = 0; i < args.size(); ++i) {
-            int val = StringTo<int>(args[i]);
+        for (auto val : counters) {
             // already enabled earlier
             if (mask & (1 << val)) {
                 std::ostringstream oss;
--- a/slsDetectorSoftware/src/CallerSpecial.cpp
+++ b/slsDetectorSoftware/src/CallerSpecial.cpp
@@ -1226,11 +1226,7 @@ std::string Caller::rx_dbitlist(int action) {
         }
         // 'none' option already covered as t is empty by default
         else if (args[0] != "none") {
-            unsigned int ntrim = args.size();
-            t.resize(ntrim);
-            for (unsigned int i = 0; i < ntrim; ++i) {
-                t[i] = StringTo<int>(args[i]);
-            }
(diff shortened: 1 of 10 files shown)


slsSupportLib/include/sls/ToString.h
```diff
@@ -407,4 +409,36 @@ std::vector<T> StringTo(const std::vector<std::string> &strings) {
...
+T StringTo(const std::string &s) {
+    using ElementType = typename T::value_type;
+    T res;
+    std::istringstream ss(s);
+    std::string item;
+    while (std::getline(ss, item, ',')) {
+        item.erase(std::remove_if(item.begin(), item.end(),
+                                  [](char c) {
+                                      return c == '[' || c == ']' || c == '"' ||
+                                             c == ' ';
```
AliceMazzoleni99, line 426: Should we throw if start of string is [ and end is not closing ] ?

erikfrojdh: We could but then we need to keep track of which symbols we found. I'm not sure it's needed but can be convinced.

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1014,15 +1014,16 @@ std::string Caller::counters(int action) {
         if (args.empty()) {
             WrongNumberOfParameters(1);
         }
-        if (std::any_of(args.cbegin(), args.cend(), [](std::string s) {
-                return (StringTo<int>(s) < 0 || StringTo<int>(s) > 2);
+        //convert args to string and then to a vector of ints
```
AliceMazzoleni99, line 1017: Update the help string in both commands - such that it includes comma seperated values and list

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1014,15 +1014,16 @@ std::string Caller::counters(int action) {
...
             })) {
             throw RuntimeError("Invalid counter indices list. Example: 0 1 2");
         }
         // convert vector to counter enable mask
         uint32_t mask = 0;
-        for (size_t i = 0; i < args.size(); ++i) {
-            int val = StringTo<int>(args[i]);
+        for (auto val : counters) {
             // already enabled earlier
             if (mask & (1 << val)) {
```
AliceMazzoleni99, line 1028: Maybe add the end use: os << ToString(counters) << '\n'; instead of args. Returns what you actually set.

slsDetectorSoftware/src/CallerSpecial.cpp
```diff
@@ -1226,11 +1227,7 @@ std::string Caller::rx_dbitlist(int action) {
...
         else if (args[0] != "none") {
-            unsigned int ntrim = args.size();
-            t.resize(ntrim);
-            for (unsigned int i = 0; i < ntrim; ++i) {
-                t[i] = StringTo<int>(args[i]);
-            }
+            t = StringTo<std::vector<int>>(ToString(args));
         }
         det->setRxDbitList(t, std::vector<int>{det_id});
         os << ToString(args) << '\n';
```
AliceMazzoleni99, line 1233: Also here ToString(t)

AliceMazzoleni99 (changes requested):
Looks good. I think there is still an issue with command trimen and threshold for mythen. Also update the Release Notes.

PR #1464, opened 2026-05-21
CI: Use conda-build instead conda build

lrlunin:
The `conda build` fails since the last update of `miniforge`. The preferred way is to use `conda-build` instead. Also updated the actions versions.


AliceMazzoleni99:
Still an issue with conda-build for newer conda versions. @lrlunin is looking into it. Gonna merge now.

AliceMazzoleni99 (approved):

PR #1468, opened 2026-05-22
Fix/fmt install

AliceMazzoleni99:
- fix python warning escaped \* in docstring 
- avoid second install of fmt


thattil (approved):
Looks good. Even got rid of the warning of cmakefile at like 218. Also python document loads properly.

Issue #1476, opened 2026-06-22
F_GET_SERVER_VERSION socket read 0 bytes


erikfrojdh:
Happened both on cSAXS which Eiger and Cristallina with JF. Probably due to different issues. 

- Improve error message 
- Monitor if it happens again. 

`DataSocket::read()` returned <= 0 — bytes_read was still 0 when the loop broke. read() returns 0 on an orderly EOF (peer closed its end of the connection) and -1 on error (ECONNRESET, timeout, etc.). In both cases you get the 0-byte message.

PR #1477, opened 2026-06-23
Better error message on socket 

erikfrojdh:
Improved error messages on TCP sockets:
- Explicitly say it's a socket error
- read errno
- add timing to see how long the faulty call took  

_A read of 0 is not a timeout and usually not an error. It means the peer’s kernel sent a TCP FIN, and your kernel delivered EOF to your application._

closing #1476

```bash
starting server on port 1951
- 14:48:42.261 ERROR: TCP socket read 10 bytes instead of 100 bytes (F_GET_SERVER_VERSION): connection closed by peer (EOF) after 0.051587 ms
starting server on port 1952
- 14:48:42.468 ERROR: TCP socket read 0 bytes instead of 100 bytes (F_EXEC_COMMAND): read error: Resource temporarily unavailable after 106.895 ms
```

### DIFF
--- a/slsSupportLib/src/ClientSocket.cpp
+++ b/slsSupportLib/src/ClientSocket.cpp
@@ -5,9 +5,11 @@
 #include "sls/sls_detector_defs.h"
 #include "sls/sls_detector_exceptions.h"
 #include "sls/sls_detector_funcs.h"
+#include "sls/string_utils.h"
 #include <arpa/inet.h>
 #include <cassert>
 #include <cstring>
+#include <fmt/format.h>
 #include <iostream>
 #include <stdexcept>
 #include <unistd.h>
@@ -24,8 +26,9 @@ ClientSocket::ClientSocket(std::string stype, const std::string &host,
     hints.ai_flags |= AI_CANONNAME;
 
     if (getaddrinfo(host.c_str(), nullptr, &hints, &result) != 0) {
-        std::string msg = socketType + " cannot decode host:" + host +
-                          " on port " + std::to_string(port) + "\n";
+
+
+        auto msg = fmt::format("Cannot resolve {} hostname: '{}'", to_lower(socketType), host);
         throwError(msg);
     }
 
@@ -40,8 +43,10 @@ ClientSocket::ClientSocket(std::string stype, const std::string &host,
     if (::connect(getSocketId(), (struct sockaddr *)&serverAddr,
                   sizeof(serverAddr)) != 0) {
         freeaddrinfo(result);
-        std::string msg = socketType + ": Cannot connect to " + host +
-                          " on port " + std::to_string(port) + "\n";
+        auto msg = fmt::format(
+            "Cannot connect to {} on {}:{}\n",
+            to_lower(socketType), host, port);
+
         throwError(msg);
     }
     freeaddrinfo(result);
@@ -53,8 +58,7 @@ ClientSocket::ClientSocket(std::string sType, struct sockaddr_in addr)
     if (::connect(getSocketId(), (struct sockaddr *)&addr, sizeof(addr)) != 0) {
         char address[INET_ADDRSTRLEN];
         inet_ntop(AF_INET, &addr.sin_addr, address, INET_ADDRSTRLEN);
-        std::string msg = socketType + ": Cannot connect to " + address +
-                          " on port " + std::to_string(addr.sin_port) + "\n";
+        auto msg = fmt::format("Cannot connect to {} on {}:{}", to_lower(socketType), address, addr.sin_port);
         throwError(msg);
     }
 }
@@ -95,7 +99,8 @@ void ClientSocket::readReply(int &ret, void *retval, size_t retval_size) {
     }
     // debugging
     catch (SocketError &e) {
-        throwError(socketType + " Socket Error: " + std::string(e.what()));
+        auto msg = fmt::format("While reading reply from {} {}", to_lower(socketType), e.what());
+        throwError(msg);
     }
 }
 
--- a/slsSupportLib/src/ClientSocket.cpp
+++ b/slsSupportLib/src/ClientSocket.cpp
@@ -24,9 +24,9 @@ ClientSocket::ClientSocket(std::string stype, const std::string &host,
     hints.ai_flags |= AI_CANONNAME;
 
     if (getaddrinfo(host.c_str(), nullptr, &hints, &result) != 0) {
-        std::string msg = "ClientSocket cannot decode host:" + host +
+        std::string msg = socketType + " cannot decode host:" + host +
                           " on port " + std::to_string(port) + "\n";
-        throw SocketError(msg);
+        throwError(msg);
     }
 
     // TODO! Erik, results could have multiple entries do we need to loop
@@ -40,10 +40,9 @@ ClientSocket::ClientSocket(std::string stype, const std::string &host,
     if (::connect(getSocketId(), (struct sockaddr *)&serverAddr,
                   sizeof(serverAddr)) != 0) {
         freeaddrinfo(result);
-        std::string msg = "ClientSocket: Cannot connect to " + socketType +
-                          ":" + host + " on port " + std::to_string(port) +
-                          "\n";
-        throw SocketError(msg);
+        std::string msg = socketType + ": Cannot connect to " + host +
+                          " on port " + std::to_string(port) + "\n";
+        throwError(msg);
     }
     freeaddrinfo(result);
 }
@@ -54,10 +53,19 @@ ClientSocket::ClientSocket(std::string sType, struct sockaddr_in addr)
     if (::connect(getSocketId(), (struct sockaddr *)&addr, sizeof(addr)) != 0) {
         char address[INET_ADDRSTRLEN];
         inet_ntop(AF_INET, &addr.sin_addr, address, INET_ADDRSTRLEN);
-        std::string msg = "ClientSocket: Cannot connect to " + socketType +
-                          ":" + address + " on port " +
-                          std::to_string(addr.sin_port) + "\n";
-        throw SocketError(msg);
+        std::string msg = socketType + ": Cannot connect to " + address +
+                          " on port " + std::to_string(addr.sin_port) + "\n";
+        throwError(msg);
+    }
+}
+
+void ClientSocket::throwError(const std::string &msg) const {
+    if (socketType == "Receiver") {
+        throw ReceiverError(msg);
+    } else if (socketType == "Detector") {
+        throw DetectorError(msg);
+    } else {
+        throw GuiError(msg);
     }
 }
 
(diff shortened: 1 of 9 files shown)


slsSupportLib/src/DataSocket.cpp
```diff
@@ -61,10 +65,16 @@ int DataSocket::Receive(void *buffer, size_t size) {
     if (bytes_read == bytes_expected) {
         return bytes_read;
     } else {
+        int err = errno; // capture before any other call can clobber it
```
thattil, line 79: If the error is EINTR, shouldnt we jus retry  to write into the socket again?

slsSupportLib/src/DataSocket.cpp
```diff
@@ -61,10 +65,16 @@ int DataSocket::Receive(void *buffer, size_t size) {
...
         std::ostringstream ss;
         ss << "TCP socket read " << bytes_read << " bytes instead of "
            << bytes_expected << " bytes ("
            << getFunctionNameFromEnum(static_cast<detFuncs>(fnum_)) << ')';
+        if (this_read == 0)
```
thattil, line 84: Could you also do this for the server side?

slsSupportLib/src/ClientSocket.cpp
```diff
@@ -40,10 +40,9 @@ ClientSocket::ClientSocket(std::string stype, const std::string &host,
     if (::connect(getSocketId(), (struct sockaddr *)&serverAddr,
                   sizeof(serverAddr)) != 0) {
         freeaddrinfo(result);
-        std::string msg = "ClientSocket: Cannot connect to " + socketType +
-                          ":" + host + " on port " + std::to_string(port) +
-                          "\n";
-        throw SocketError(msg);
+        std::string msg = socketType + ": Cannot connect to " + host +
```
AliceMazzoleni99: Isnt it easier for debugging if info ClientSocket remains? For me the error reads more as if detector server tries to connect to host

AliceMazzoleni99: i mean its good to add the socket type e.g. to which server to connect.

AliceMazzoleni99: maybe better: fmt::format("ClientSocket: Cannot connect to {} with hostname {} on port {}\n", sockettype, host, std::to_string(port));

slsSupportLib/src/ClientSocket.cpp
```diff
@@ -80,26 +88,14 @@ void ClientSocket::readReply(int &ret, void *retval, size_t retval_size) {
...
     // debugging
     catch (SocketError &e) {
-        if (socketType == "Receiver") {
-            throw ReceiverError("Receiver returned: " + std::string(e.what()));
-        } else if (socketType == "Detector") {
-            throw DetectorError("Detector returned: " + std::string(e.what()));
-        } else {
-            throw GuiError(e.what());
-        }
+        throwError(socketType + " Socket Error: " + std::string(e.what()));
```
AliceMazzoleni99: maybe also more explicit: throwError(fmt::format("Socket Error when reading from {}: {}", socketType, std::string(e.what())));

slsSupportLib/src/DataSocket.cpp
```diff
@@ -61,10 +65,16 @@ int DataSocket::Receive(void *buffer, size_t size) {
...
+            ss << ": connection closed by peer (EOF)";
+        else if (this_read < 0)
+            ss << ": read error:  " << std::strerror(err) << " (" << errno_name(err) << ")";
+        ss << " after " << timer.elapsed_ms() << " ms";
         throw SocketError(ss.str());
     }
 }
```
AliceMazzoleni99, line 92: maybe also distinguish error for send

slsSupportLib/src/DataSocket.cpp
```diff
@@ -61,10 +65,16 @@ int DataSocket::Receive(void *buffer, size_t size) {
...
     } else {
+        int err = errno; // capture before any other call can clobber it
         std::ostringstream ss;
         ss << "TCP socket read " << bytes_read << " bytes instead of "
            << bytes_expected << " bytes ("
            << getFunctionNameFromEnum(static_cast<detFuncs>(fnum_)) << ')';
+        if (this_read == 0)
+            ss << ": connection closed by peer (EOF)";
+        else if (this_read < 0)
+            ss << ": read error:  " << std::strerror(err) << " (" << errno_name(err) << ")";
```
AliceMazzoleni99: As Dhanya mentioned maybe its good to check if ignore version compatibility was set. And change the error message respectively.

slsSupportLib/tests/test-Sockets.cpp
```diff
@@ -50,4 +72,93 @@ TEST_CASE("throws on no server", "[support]") {
...
+
+    auto s = std::async(std::launch::async, echo_server, 1960, 100,
+                        std::chrono::milliseconds(0));
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    auto client = DetectorSocket("localhost", 1960);
+    client.Send(sent_message.data(), sent_message.size());
+    
+
+    REQUIRE_THROWS(client.Receive(received_message.data(), received_message.size()));
+    // client.close();
```
AliceMazzoleni99: why not closing?

AliceMazzoleni99 (commented):
Added some minor comments. Error handling si for sure better. Also on server side we only check for read returns  == 0.

slsSupportLib/src/ClientSocket.cpp
erikfrojdh: Now reads

```bash
- 16:28:37.346 ERROR: Cannot connect to detector on localhost:1950
- 16:28:38.178 ERROR: Cannot resolve detector hostname: 'invalidhostname'
- 16:47:33.650 ERROR: While reading reply from detector TCP socket read 1 bytes instead of 4 bytes (F_GET_DETECTOR_TYPE): connection closed by peer (EOF) after 0.023542 ms
```

slsSupportLib/src/DataSocket.cpp
thattil: Bu this I can do a separate PR afterwards for this if its too much here.

slsSupportLib/tests/test-Sockets.cpp
```diff
@@ -154,11 +193,78 @@ TEST_CASE("Socket crash?", "[support]") {
...
+        error_message = e.what();
+    }
+    client.close();
+    s.get();
+
+    // The client read only 1 of the 4 expected retval bytes and then hit EOF.
+    CHECK_THAT(error_message,
+               Catch::Matchers::Contains("read 1 bytes instead of 4 bytes"));
+    CHECK_THAT(error_message,
+               Catch::Matchers::Contains("connection closed by peer (EOF)"));
```
AliceMazzoleni99, line 288: much cleaner code.

slsSupportLib/src/ClientSocket.cpp
erikfrojdh: message is now: 
- 19:21:23.915 ERROR:  While reading reply from detector TCP socket read 1 bytes instead of 4 bytes (F_GET_DETECTOR_TYPE): connection closed by peer (EOF) after 0.023625 ms

slsSupportLib/src/DataSocket.cpp
erikfrojdh: I think we should do this in another PR. Not sure where to check since at the moment the client socket is not aware of the SHM and I'm not so keen in introducing a dependency. If we do it probably we should find a place where we can catch the error and insert that info... if possible.

slsSupportLib/src/DataSocket.cpp
erikfrojdh: Do we have any use case in our code where we want to use a signal handler and then continue? We use SIGTERM on the receiver but then we don't want to try again...

slsSupportLib/src/DataSocket.cpp
erikfrojdh: Added a check on read as well.

slsSupportLib/src/DataSocket.cpp
erikfrojdh: Done

slsSupportLib/tests/test-Sockets.cpp
erikfrojdh: Close is back in, but in principle not needed since we close in the destructor. However, we need to expose close to allow for closing the socket at a specific time.

slsSupportLib/src/DataSocket.cpp
erikfrojdh: Added retry on EINTR. Likely not needed but good practice

erikfrojdh:
Found two issues which we can also fix while working on the socket class. Question do we want to do them in this PR or in a separate one?

**Misleading timeout functions**
```cpp
// Always sets receive timeout to indefinite (0) before setting send timeout
// Reads error code and throws on error
int DataSocket::setTimeOut(int t_seconds);

// Returns setsocketopt's return code directly without checking
int DataSocket::setReceiveTimeout(int us);
```

**Send/Receive returns number of bytes as int**

- Fixed internally to use size_t/ssize_t to avoid truncation 
- For now keeping the API (guess we are not sending more than 2GB...)
- Better would be to return ssize_t/size_t or if we always throw on fail just make them void

**Redundant read/write**

We expose read/write on the underlying socket should probably remove then and only use Send/Receive?

slsSupportLib/src/DataSocket.cpp
```diff
@@ -23,6 +23,14 @@ namespace sls {
 DataSocket::DataSocket(int socketId) : sockfd_(socketId) {
     int value = 1;
     setsockopt(sockfd_, SOL_SOCKET, SO_REUSEADDR, &value, sizeof(value));
+#ifdef SO_NOSIGPIPE
+    // macOS/BSD: suppress SIGPIPE when sending to a peer that closed the
+    // connection, so a failed send returns an error instead of killing the
+    // process. On Linux we instead pass MSG_NOSIGNAL to send() (see Send()).
+    int nosigpipe = 1;
+    setsockopt(sockfd_, SOL_SOCKET, SO_NOSIGPIPE, &nosigpipe,
```
AliceMazzoleni99, line 31: maybe also check for errors in setsockopt.

AliceMazzoleni99: Minor: Should SIGPIPE handler be removed in MatterhornServerApp, ReceiverApp etc.

AliceMazzoleni99 (approved):

AliceMazzoleni99:
> Found two issues which we can also fix while working on the socket class. Question do we want to do them in this PR or in a separate one?
> 
> **Misleading timeout functions**
> 
> ```c++
> // Always sets receive timeout to indefinite (0) before setting send timeout
> // Reads error code and throws on error
> int DataSocket::setTimeOut(int t_seconds);
> 
> // Returns setsocketopt's return code directly without checking
> int DataSocket::setReceiveTimeout(int us);
> ```
> 
> **Send/Receive returns number of bytes as int**
> 
>     * Fixed internally to use size_t/ssize_t to avoid truncation
> 
>     * For now keeping the API (guess we are not sending more than 2GB...)
> 
>     * Better would be to return ssize_t/size_t or if we always throw on fail just make them void
> 
> 
> **Redundant read/write**
> 
> We expose read/write on the underlying socket should probably remove then and only use Send/Receive? 

I would remove read/write in this PR and only use send/receive. For setTimeOut it does not throw it just logs the error. Maybe throwing an error is better. Also maybe use setReciverTimeOut in setTimeOut makes it a bit more clear.

slsSupportLib/src/DataSocket.cpp
thattil: ill figure it out in another PR. all good.

erikfrojdh:
Removed read/write that were almost not used.

slsSupportLib/src/DataSocket.cpp
erikfrojdh: I think we can remove the SIGPIPE check since the sockets shouldn't fail with that now, but would be good to verify. Making an issue...

PR #1480, opened 2026-06-25
dev: zmq hwm rebind

thattil:
- move hwm into the zmq constructor and reconstruct the socket instead of rebind. seems to work better than rebind (most connections cant bind so fast?)
- added test to reconnect zmq sockets (20 eiger half mods) for zmq port and zmq hwm set up
- changed default ports of detector adn receiver tcp ports to be able to handle 20 connections.


erikfrojdh:
```
int zmq_setsockopt (void *socket, int option_name, const void *option_value, size_t option_len);
```

_Caution: All options, with the exception of ZMQ_SUBSCRIBE, ZMQ_UNSUBSCRIBE and ZMQ_LINGER, only take effect for subsequent socket bind/connects._

Setting before binding works well the first time. What happens when the user wants to modify the hwm later?

thattil:
> ```
> int zmq_setsockopt (void *socket, int option_name, const void *option_value, size_t option_len);
> ```
> 
> _Caution: All options, with the exception of ZMQ_SUBSCRIBE, ZMQ_UNSUBSCRIBE and ZMQ_LINGER, only take effect for subsequent socket bind/connects._
> 
> Setting before binding works well the first time. What happens when the user wants to modify the hwm later?

Initially, we were doing rebind (custom function that calls unbind and then bind). This doesnt seem to work for fedora or some environments. So, now I am implementing hwm the same way as zmq port. After the socket was created, if one wants to change it, it is destroyed, port is set and socket created again.  Therefore, I am repeating the same procedure for hwm. Destroying, set it and creating the sockets again. If you have another option, please let me know.


And for the SetHWM function, it is written in the header that it requires a rebind after.

AliceMazzoleni99 (commented):
Looks good to me. I think Release notes should be updated.

AliceMazzoleni99 (approved):

PR #1481, opened 2026-06-26
added moench05 to pyctbgui

AliceMazzoleni99:
- add moench 05 transform to pyctbgui

### DIFF
--- a/pyctbgui/pyctbgui/services/Plot.py
+++ b/pyctbgui/pyctbgui/services/Plot.py
@@ -7,7 +7,7 @@ import numpy as np
 from PyQt5 import QtWidgets, QtGui, QtCore, uic
 
 from aare import transform, ReadoutMode 
-from aare._aare import Matterhorn10, Matterhorn02, Moench04
+from aare._aare import Matterhorn10, Matterhorn02, Moench04, Moench05
 
 import pyqtgraph as pg
 from pyctbgui.utils import recordOrApplyPedestal 
@@ -530,8 +530,8 @@ class PlotTab(QtWidgets.QWidget):
             self.mainWindow.nAnalogCols = Moench04.nCols
             self.mainWindow.decoder = transform.Moench04AnalogTransform() 
         elif self.view.comboBoxPlot.currentText() == "Moench05":
-            self.mainWindow.nAnalogRows = 160
-            self.mainWindow.nAnalogCols = 150
+            self.mainWindow.nAnalogRows = Moench05.nRows
+            self.mainWindow.nAnalogCols = Moench05.nCols 
             self.mainWindow.decoder = transform.Moench05Transform()
 
         try: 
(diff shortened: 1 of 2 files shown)


pyctbgui/pyctbgui/services/Plot.py
```diff
@@ -529,6 +529,10 @@ def setDecoder(self):
             self.mainWindow.nAnalogRows = Moench04.nRows
             self.mainWindow.nAnalogCols = Moench04.nCols
             self.mainWindow.decoder = transform.Moench04AnalogTransform() 
+        elif self.view.comboBoxPlot.currentText() == "Moench05":
+            self.mainWindow.nAnalogRows = 160
```
thattil: I would rather put these hard coded definitions in defines.py

thattil (approved):
Just minor changes. LGTM

mmarti04:
how do you know which adc's to look at ? Or is this a problem for aare ?

pyctbgui/pyctbgui/services/Plot.py
AliceMazzoleni99: I aggree. Just wanted to avoid doing another PR in aare. I will add the info in aare. Makes more sense to add chip specific stuff only used for transform there.

AliceMazzoleni99:
> how do you know which adc's to look at ?

The user has to set it. I can create a ticket for proper error handling if the wrong adcs are set. And proper documentation of the transform.

mmarti04:
> > how do you know which adc's to look at ?
> 
> The user has to set it. I can create a ticket for proper error handling if the wrong adcs are set. And proper documentation of the transform.

yes but the user cannot select the ADC's that are needed for moench when 10g is enabled. For the newest adapter for example we would need ADC's 1,9 and 13. Without 10g the user can enable exactly these, but with 10g we will instead enable ADC's 0,1,2,3,8,9,10,11,12,13,14 and 15. How does the software know which ones of that are the ones used by Moench ?

pyctbgui/pyctbgui/services/Plot.py
AliceMazzoleni99: I agree. Added it in aare: https://github.com/slsdetectorgroup/aare/pull/333

AliceMazzoleni99:
> > > how do you know which adc's to look at ?
> > 
> > 
> > The user has to set it. I can create a ticket for proper error handling if the wrong adcs are set. And proper documentation of the transform.
> 
> yes but the user cannot select the ADC's that are needed for moench when 10g is enabled. For the newest adapter for example we would need ADC's 1,9 and 13. Without 10g the user can enable exactly these, but with 10g we will instead enable ADC's 0,1,2,3,8,9,10,11,12,13,14 and 15. How does the software know which ones of that are the ones used by Moench ?

This is taken care of in the decoder. So in the decoder it uses ADC's 5,9,1.

thattil (approved):
👍

PR #1484, opened 2026-07-07
Dev/fix workflow cmake cache

thattil:
- Add checks for cached apt dependencies to catch incomplete installations. 
- A temporary Ubuntu mirror/package index mismatch caused apt installation to fail, but the failure was not detected before CMake ran.
- awalsh128/cache-apt-pkgs-action@v1.6.1 (instead of latest): issue: https://github.com/awalsh128/cache-apt-pkgs-action/issues/218 
- deleted all the caches to start with fresh caches.


AliceMazzoleni99 (approved):
Looks good.

PR #1485, opened 2026-07-07
fix workflow cmake cache 2

thattil:
- remnants of #1484  
- removed the verification of the packages
- added the fixed version v1.6.1 instead of latest for the awalsh128/cache-apt-pkgs-action repo to create/restore cache.


.github/workflows/run_tests.yaml
```diff
@@ -24,33 +24,9 @@ jobs:
       - uses: awalsh128/cache-apt-pkgs-action@v1.6.1
         with:
           packages: libhdf5-dev libfmt-dev
+          #packages: libfmt-dev
```
AliceMazzoleni99: maybe remove.

AliceMazzoleni99 (approved):
Looks good to me.

PR #1488, opened 2026-07-10
generic HDF5 filters & compression

### LINKED ISSUE #1235: hdf5: compress and bitshuffle filter

mmarti04:
Adds the option to use a generic hdf5 compression filter. 
Should probably do this as a command at some point, for the moment i added it at compile time. 

Example: gzip-deflate, lvl 9 deflate compression: 
```
cmake .. -DSLS_USE_HDF5=ON -DSLS_HDF5_FILTER_ID=1 -DSLS_HDF5_FILTER_PARAMS="9"
...
p fformat hdf5
acquire
```

Only these filter-ID's are compiled directly into HDF5:
| Filter ID | Name                    | Description                                                        |
|-----------|-------------------------|--------------------------------------------------------------------|
| 0         | `H5Z_FILTER_NONE`       | No filter                                                          |
| 1         | `H5Z_FILTER_DEFLATE`    | GZIP / zlib deflate compression                                    |
| 2         | `H5Z_FILTER_SHUFFLE`    | Byte shuffle (reorders bytes to improve compression)               |
| 3         | `H5Z_FILTER_FLETCHER32` | Fletcher32 checksum (error detection)                              |
| 4         | `H5Z_FILTER_SZIP`       | SZIP compression                                                   |
| 5         | `H5Z_FILTER_NBIT`       | N-bit compression (for integer/float data with unused bits)        |
| 6         | `H5Z_FILTER_SCALEOFFSET`| Scale and offset compression (lossy for floats, lossless for ints) |

All others (bitshuffle, lz4, my on-FPGA compression etc.) need to be compiled externally and are dynamically registered by setting HDF5_PLUGIN_PATH. 

[Link HDF5 filters](https://github.com/HDFGroup/hdf5_plugins/blob/master/docs/RegisteredFilterPlugins.md)

My custom FPGA-compression hdf5 filter is only working in toy-examples right now but i think this might be useful even without the fpga side.


mmarti04:
Also briefly tested with bitshuffle+lz4 now. 

- One needs to compile this: [Bitshuffle Github](https://github.com/kiyo-masui/bitshuffle), (src, not python !)
- Set HDF5_PLUGIN_PATH to the install directory. 
- Then take the correct ID and config values from here: [filter Info](https://github.com/HDFGroup/hdf5_plugins/blob/master/docs/RegisteredFilterPlugins.md#bitshuffle)
- and set them with cmake:
 ```
cmake .. -DSLS_USE_HDF5=ON -DSLS_HDF5_FILTER_ID=32008 -DSLS_HDF5_FILTER_PARAMS="0,2"
```

seems to work

File: slsReceiverSoftware/src/HDF5DataFile.cpp [hunk 1]
```diff
@@ -4,6 +4,15 @@
 #include "receiver_defs.h"
 
 #include <iomanip>
+#include <sstream>
+
+#ifndef SLS_HDF5_FILTER_ID
+#define SLS_HDF5_FILTER_ID 0
+#endif
+
```
AliceMazzoleni99: Is that neccessary? There are default values set in cmake.

slsReceiverSoftware/src/HDF5DataFile.cpp
```diff
@@ -193,6 +202,37 @@ void HDF5DataFile::CreateFile() {
...
+            const auto filterId = static_cast<H5Z_filter_t>(SLS_HDF5_FILTER_ID);
+            if (H5Zfilter_avail(filterId) <= 0) {
+                throw RuntimeError("HDF5 filter ID " + std::to_string(SLS_HDF5_FILTER_ID) + " is not available, check HDF5_PLUGIN_PATH");
+            }
+
+            // parse generic filter parameters
+            std::vector<unsigned int> filterParams;
+            std::stringstream ss(SLS_HDF5_FILTER_PARAMS);
+            std::string token;
+            while (std::getline(ss, token, ',')) {
```
AliceMazzoleni99: I think it would have been cleaner to use a list directly e.g. in HDF5DataFIle.hpp use constexpr std::array filterParams = {SLS_HDF5_FILTER_PARAMS}. But as a command generally makes more sense I dont think we need it.

AliceMazzoleni99 (commented):
Some minor comments. For me it definitely makes more sense to have a command to set the filter. I dont quite understand the HDF5_PLUGIN_PATH was it considered in the cmake?

mmarti04:
> Some minor comments. For me it definitely makes more sense to have a command to set the filter. I dont quite understand the HDF5_PLUGIN_PATH was it considered in the cmake?

@AliceMazzoleni99 What do you mean with "considered in cmake" ? 
It's pointing to the shared library path of the filter plugins. They are dynamically loaded at runtime. Why would we need them in cmake ?

slsReceiverSoftware/src/HDF5DataFile.cpp
mmarti04: removed.

slsReceiverSoftware/src/HDF5DataFile.cpp
mmarti04: while we still have it as a cmake option yes, once it becomes a command it should be a vector i guess so i would leave it for now.

AliceMazzoleni99 (approved):

PR #1491, opened 2026-07-14
Fix rx_roi and metadata tcp communication 

thattil:
added sending ok to client from receiver for a get roi/metadata before sending the retvals because an exception before will not be sent

### DIFF
--- a/slsReceiverSoftware/src/ClientInterface.cpp
+++ b/slsReceiverSoftware/src/ClientInterface.cpp
@@ -1740,8 +1740,7 @@ int ClientInterface::get_receiver_roi(Interface &socket) {
                            std::to_string(size) + ". Expected: " +
                            std::to_string(impl()->getNumberofUDPInterfaces()));
     }
-    socket.Send(OK);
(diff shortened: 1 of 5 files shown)


slsReceiverSoftware/src/ClientInterface.cpp
```diff
@@ -1740,6 +1740,7 @@ int ClientInterface::get_receiver_roi(Interface &socket) {
                            std::to_string(size) + ". Expected: " +
                            std::to_string(impl()->getNumberofUDPInterfaces()));
     }
+    socket.Send(OK);
     socket.Send(size);
```
AliceMazzoleni99: I think better is to use socket.sendResult on receiver side and client.readReply on client side.

AliceMazzoleni99 (commented):
I would use sendResult and readReply instead of handling FAIL and send OK manually in the code. Does one need to update the release notes?

thattil:
Closing it as another PR is done and much more apt. #1492

PR #1492, opened 2026-07-20
tcp: Variable size array

### LINKED ISSUE #1460: rx get roi tcp
rx roi tcp does not send the ok/fail before sending roi (similarly on client side). the errors  on rx side will not make sense without the ok/ fail being sent first

get rx roi
get rx roi metadata

thattil:
Tcp variable vector size needs to send size first. Abstractign this from Module in client and ClientInterface class in receiver. 

If this works, can do for the others.

### DIFF
--- a/slsDetectorSoftware/src/Module.cpp
+++ b/slsDetectorSoftware/src/Module.cpp
@@ -2005,50 +2005,41 @@ void Module::sendVetoPhoton(const int chipIndex,
                             const std::vector<int> &gainIndices,
                             const std::vector<int> &values) {
     const int nch = gainIndices.size();
-    if (gainIndices.size() != values.size()) {
-        throw RuntimeError("Number of Gain Indices and values do not match! "
-                           "Gain Indices size: " +
-                           std::to_string(gainIndices.size()) +
-                           ", values size: " + std::to_string(values.size()));
+    if (nch != static_cast<int>(values.size())) {
+        std::ostringstream oss;
+        oss << "Number of Gain Indices and values do not match! Gain Indices "
+               "size: "
+            << nch << ", values size: " << values.size();
+        throw RuntimeError(oss.str());
     }
     LOG(logDEBUG1) << "Sending veto photon/file to detector [chip:" << chipIndex
                    << ", nch:" << nch << "]";
-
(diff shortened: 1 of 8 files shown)


slsDetectorSoftware/src/Module.cpp
```diff
@@ -3379,6 +3356,44 @@ Ret Module::sendToDetectorStop(int fnum, const Arg &args) {
...
+        oss << getFunctionNameFromEnum(static_cast<detFuncs>(fnum));
+        throw RuntimeError(oss.str());
+    }
+    auto receiver = ReceiverSocket(shm()->rxHostname, shm()->rxTCPPort);
+    receiver.sendCommandVariableSize(fnum, args, nullptr, 0);
+    receiver.close();
+}
+
+template <typename Ret>
+std::vector<Ret> Module::sendToReceiverVarVector(int fnum) const {
```
AliceMazzoleni99, line 3205: Is maybe receiveFromReceiverVarVector a better name?

slsDetectorSoftware/src/Module.cpp
```diff
@@ -3379,6 +3356,44 @@ Ret Module::sendToDetectorStop(int fnum, const Arg &args) {
...
+                   << "]";
+    if (!shm()->useReceiverFlag) {
+        std::ostringstream oss;
+        oss << "Set rx_hostname first to use receiver parameters, ";
+        oss << getFunctionNameFromEnum(static_cast<detFuncs>(fnum));
+        throw RuntimeError(oss.str());
+    }
+    std::vector<Ret> retval;
+    auto receiver = ReceiverSocket(shm()->rxHostname, shm()->rxTCPPort);
+    receiver.sendCommandVariableSize(fnum, nullptr, 0, retval);
```
AliceMazzoleni99, line 3218: Maybe also better sendCommandReceiveVariableSize

AliceMazzoleni99 (approved):
Much cleaner. I would definitly use this. The naming is a bit odd for me  e.g. hard to know whats going on - but I see the consistency with sendToReceiver. So actually we can stick to it.

AliceMazzoleni99 (approved):

PR #1495, opened 2026-07-27
1001/fix conda build

thattil:


thattil:
Decided to do each consecutive PR on top of 10.1.0.rc instead of trying to fix 10.0.1.rc which will not be fixed consecutive in time.

PR #1448, opened 2026-05-07
Dev/rx disable datastream port

thattil:
- changed 
   - command from `datastream` to `udp_datastream` to be more specific
   - Detector API:
      - setNumberofUDPInterfaces(pos) =>setNumberofUDPInterfaces() # no position anymore
      - getDataStream(portPosition, pos) =>getUDPDataStream(portPosition, pos)
      - setDataStream(portPosition, pos) => setUDPDataStream(portPosition) # no position anymore
      
   - modified the dataprocessing thread: 
      - if gui or call back, start zmq processing (connecting sockets) before starting receiver (disabled ports sends dummy that the client zmq sockets are not ready for)
      - if only progress, starting progress processing thread after startReceiver
      -

- added
   - Detector API:
      - getRxDisabledUDPPortIndices()
      - getPortPositionList()  
   - allow disabling/enabling udp interface in receiver for Jungfrua/Moench when both udp interfaces is enabled  (This is to go fast, where only one half of the module matters)
   - write ports type and disabled ports to master file (tests for it as well)
   - tests for python commands as well
   - datastream command to deprecated commands
   - refactored tests to make the new parameters fit

PR #1497, opened 2026-07-27
enabled data processor should write to master file

AliceMazzoleni99:
- previously DataProcess 0 created master file 
- if disabled information about caught frames is wrong and screws up file reading in aare 
- This PR fixes: https://github.com/slsdetectorgroup/slsDetectorPackage/pull/1448

### DIFF
--- a/slsReceiverSoftware/src/Implementation.cpp
+++ b/slsReceiverSoftware/src/Implementation.cpp
@@ -1015,8 +1015,14 @@ void Implementation::StartMasterWriter() {
             masterAttributes.udpPortsDisabled = udpPortsDisabledMetadata;
 
             // create master file
-            size_t dataprocessor_index =
-                dataProcessor[0]->GetDataStreamEnable() ? 0 : 1;
+            size_t dataprocessor_index = udpDataStream[0] ? 0 : 1;
+
+            LOG(logDEBUG) << "dataprocessor_index: " << dataprocessor_index
+                          << " dataProcessor.size(): " << dataProcessor.size()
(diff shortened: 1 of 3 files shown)


slsReceiverSoftware/src/Implementation.cpp
```diff
@@ -1013,9 +1015,12 @@ void Implementation::StartMasterWriter() {
             masterAttributes.udpPortsDisabled = udpPortsDisabledMetadata;
 
             // create master file
-            masterFileName = dataProcessor[0]->CreateMasterFile(
-                filePath, fileName, fileIndex, overwriteEnable, silentMode,
-                fileFormatType, &masterAttributes, &hdf5LibMutex);
+            size_t dataprocessor_index = udpDataStream[0] ? 0 : 1;
```
thattil, line 1018: what if there arent 2 interfaces enabled? what if both the ports do not have roi.and StartMasterWriter is based on moduleposition 0 and not if they are in roi.. Maybe this should also be updated?

slsReceiverSoftware/src/Implementation.cpp
```diff
@@ -1013,9 +1015,12 @@ void Implementation::StartMasterWriter() {
...
+
+            masterFileName =
+                dataProcessor[dataprocessor_index]->CreateMasterFile(
```
thattil, line 1021: Not just for master file, but also for create virtual file, and link in masterfile..

thattil (changes requested):
Not sure what is the best way forward, but there are some areas to cover..

AliceMazzoleni99:
Handled directly in aare. This PR is deprecated.

PR #1516, opened 2026-08-10
gated Matterhorn on XCTB

mmarti04:
adds gate and trigger options to xilinx chiptestboard for firmware version fb3c8a66 and later.

Instructions and example: https://gitea.psi.ch/detectors/chiptesting/src/branch/master/Xilinx_CTB_examples/gate_and_trigger


erikfrojdh (approved):

thattil:
@mmarti04 shouldnt you also update the firmware version?

PR #1525, opened 2026-08-14
Dev/chip version string

### LINKED ISSUE #1493: Jungfrau chip v1.2
For 10.1.0 Jungfrau:
- server side still reads chipversion as int from server config file (10, 11, 120, 121, 122)
- (returns double) getChipVersion 1.0, 1.1, 1.2
- (returns string) getChipVersion: 1.2 Normal, 1.2 Low Noise, 1.2 HDR, 1.3 Burst Mode
 
- 1.2.0:
   - DACS: change for every chip, setting (HG0, G0), electron collection mode.

Change for developer (11.0.0):
- remove getChipVersion returning double

thattil:
- changed api from double to string return for getchipversion, 
- updated python bindings, 
- added to versions command, 
- fixed appropriately to send string from server, 
- fixed tests accordingly to convert the string to int to do the tests


slsDetectorSoftware/src/Caller.cpp
```diff
@@ -1427,7 +1427,7 @@ std::string Caller::chipversion(int action) {
     // print help
     if (action == slsDetectorDefs::HELP_ACTION) {
         os << R"V0G0N(
-	[Jungfrau] Returns chip version. Can be 1.0 or 1.1 )V0G0N"
+	[Jungfrau] Returns chip version. )V0G0N"
```
erikfrojdh: Maybe note as string and add one example?

erikfrojdh (approved):
only one minor comment

PR #1540, opened 2026-08-24
Dev/pr automation part 1

thattil:
## PR type

**Select exactly one:**

- [ ] Feature
- [ ] Bug Fix
- [x] Infrastructure

## Breaking API

- [ ] This is a Breaking API change (Feature or Bug Fix).

## Description

It creates 3 workflows:
1. Pr validation:
   - It goes through the PR description and validates the PR primary type (feature, bug fix or infrastructure)
   - checks if breaking API is checked
   - ensures breaking api and infrastructure are not both checked
   - creates the labels for PR type and if breaking api is checked
   - trigger the PR milestone workflow
  
2. PR milestone trigger:
- If PR type or breaking api labels are changed, it will trigger the PR milestone workflow.

3. PR milestone:
- Again ensure breaking api and infrastructure are not both labelled
- reads the release_target.yaml file and get the milestones
- ensure these milestones actually exist in repo
- set according milestones depending on PR type and breaking api
   - breaking api: major milestone
   - feature and infrastructure: minor milestone
   - bugfix: bug fix milestone

PR #1542, opened 2026-08-25
Dev/pr automation part 2

thattil:
## PR type

**Select exactly one:**

- [ ] Feature
- [ ] Bug Fix
- [x] Infrastructure

## Breaking API

- [ ] This is a Breaking API change (Feature or Bug Fix).

## Description

Automaitcally add detector labels to PRs. (ignore for infrastructure PRs)
- match whole word of detector names in 
   - server folders, 
   - matching in pr title description, 
   - linked or references issues and prs, 
   - change in code with matching labels, 
   - add detector labels.
 

## Equivalent PRs

#1540


.github/workflows/pr_detector_labels.yaml
```diff
@@ -0,0 +1,250 @@
...
+             * 3. Detect detector from changed code
+             * --------------------------------------------------
+             *
+             * Only the PR patch is searched.
+             *
+             * Therefore, text that already existed elsewhere
+             * in an unchanged file does not trigger a label.
+             */
+            for (const file of files) {
+              if (!file.patch) {
```
AliceMazzoleni99, line 168: Is that neccessary? Would it be in changed file?

.github/workflows/pr_detector_labels.yaml
```diff
@@ -0,0 +1,250 @@
...
+             * --------------------------------------------------
+             * 3. Detect detector from changed code
+             * --------------------------------------------------
+             *
+             * Only the PR patch is searched.
+             *
+             * Therefore, text that already existed elsewhere
+             * in an unchanged file does not trigger a label.
+             */
+            for (const file of files) {
```
AliceMazzoleni99: Would you consider collapsing the two file loops ?

.github/workflows/pr_detector_labels.yaml
```diff
@@ -0,0 +1,250 @@
...
+              const match = file.filename.match(
+                /^slsDetectorServers\/([^/]+)DetectorServer\//
+              );
+
+              if (!match) {
+                continue;
+              }
+
+              const detectorName = match[1];
```
AliceMazzoleni99: I dotn know if we should also include stuff in commonServerfunctions.c

AliceMazzoleni99 (commented):
Looks really good. Lets see how it works in practice. The only thing if we change a documentation or comment about a detector or format -> it will already detect it. But I guess we just have to leave with it and remove optionally.

.github/workflows/pr_detector_labels.yaml
thattil: Yes, this is what we discussed about the client side. If the changed code includes an enum or even a comment, then this would be in the changed files.

.github/workflows/pr_detector_labels.yaml
thattil: Well, if anything changed there that should relate to a detector, the detector server would anyway be recompiled.

.github/workflows/pr_detector_labels.yaml
thattil: Yes will try to combine.

.github/workflows/pr_detector_labels.yaml
thattil: ah i misunderstood. it seems like a sanity check.

AliceMazzoleni99 (approved):

